PR #19, opened 2026-05-12
Honor unsafe_flag_value_prefix in shell rule evaluation

### LINKED ISSUE #15: Honor `unsafe_flag_value_prefix` in shell rule evaluation
Follow-up to #14.

## Problem

The shell rule data already declares `unsafe_flag_value_prefix` for cases that should never be auto-allowed, but the classifier does not currently enforce that field.

Current examples from `rules/shell.json`:

- `find -exec`, `-execdir`, `-ok`, `-okdir`
- `gh api --input`

That means mutating commands can currently classify as `safe` even though the rule file says otherwise.

## Reproductions

```bash
find . -name '*.py' -exec rm {} \;
gh api /repos/x/y/issues --input body.json
```

## Root cause

`rules/shell.json` expresses the policy, but `check_unsafe_flags()` only evaluates:

- `unsafe_flag_values`
- `unsafe_flag_any_value`
- `unsafe_flags_without_value`

It ignores `unsafe_flag_value_prefix`, so the implementation and rule schema have drifted.

## Design

1. Extend `check_unsafe_flags()` to evaluate prefix/value-carrying unsafe flag patterns.
2. Keep the rule shape simple for this release:
   - exact flag name match
   - value present either as `--flag=value` or `--flag value`
   - prefix / wildcard match against the value
3. Return a human-readable explanation that includes both the flag and the matched value.
4. Add rule-schema validation at load time so unknown or unimplemented keys fail loudly in tests instead of silently being ignored.

## Suggested implementation notes

- Keep the current `check_unsafe_flags()` API shape and extend it rather than introducing a parallel mechanism.
- Validate schema once during rules load, not on every classification.
- Fail closed in tests: if rule data references an unsupported field, CI should fail.

## Tests

Add regression coverage for:

- `find . -exec rm {} \;`
- `find . -execdir rm {} \;`
- `find . -ok rm {} \;`
- `gh api ... --input body.json`
- inline `--flag=value` and split `--flag value` forms

## Acceptance criteria

- The reproductions above no longer classify as `safe`.
- Rules using `unsafe_flag_value_prefix` are exercised by tests.
- New rule-schema drift is caught in CI instead of silently ignored.

debedb:
## Summary

- Extend `check_unsafe_flags()` in `hooks/rule_classifier.py` to honor `unsafe_flag_value_prefix`. The rule data already declared this field for `find -exec / -execdir / -ok / -okdir` and `gh api --input`, but the implementation only evaluated `unsafe_flag_values`, `unsafe_flag_any_value`, and `unsafe_flags_without_value`. Both split (`--flag value`) and inline (`--flag=value`) forms are matched; the value is fnmatch'd against the per-flag glob.
- Add `validate_shell_rules()` for load-time schema drift detection. A new `TestValidateShellRules` suite asserts the bundled `rules/shell.json` validates clean and that unknown top-level keys, command keys, defaults, nested subcommand keys, service-override keys, interpreter keys, and nested module keys are all flagged.
- Add regression coverage in `tests/test_grammar_classifier.py` for the four `find` exec-family flags and `gh api --input` in both forms.

After this change:

```
$ python3 hooks/grammar_classifier.py "find . -name '*.py' -exec rm {} \;"
{"decision": "unsafe", "reason": "find: flag -exec rm"}
$ python3 hooks/grammar_classifier.py 'gh api /repos/x/y/issues --input body.json'
{"decision": "unsafe", "reason": "gh api: flag --input body.json"}
```

Closes #15.

## Test plan

- [x] `python3 -m unittest discover -v tests` — 237 tests pass
- [x] CLI repros classify as `unsafe`:
  - [x] `find . -name '*.py' -exec rm {} \;`
  - [x] `find . -name '*.py' -execdir rm {} \;`
  - [x] `find . -name '*.py' -ok rm {} \;`
  - [x] `find . -name '*.py' -okdir rm {} \;`
  - [x] `gh api /repos/x/y/issues --input body.json`
  - [x] `gh api /repos/x/y/issues --input=body.json`
- [x] `validate_shell_rules(load_shell_rules(REPO_ROOT / "rules"))` returns `[]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,10 +410,13 @@ YOLT logs every examined Bash invocation by default to
 {"ts": "2026-05-08T14:00:00.000+00:00", "decision": "safe", "reason": "ls: read-only", "command": "ls /tmp"}
 ```
 
-`decision` is one of `safe`, `unsafe`, `unknown`, or `import-error` (the
-last when the tree-sitter dependency is missing). The `command` field is
-truncated to 500 characters. Logging failures are swallowed — the hook
-never breaks the session because of an unwritable log path.
+`decision` is one of `safe`, `unsafe`, `unknown`, `import-error` (the
+tree-sitter dependency is missing), or `rules-validation-error` (the
+bundled or user-override `shell.json` failed schema validation; the
+`reason` field carries the list of offending keys / defaults so the
+user can fix the override). The `command` field is truncated to 500
+characters. Logging failures are swallowed — the hook never breaks the
+session because of an unwritable log path.
 
 ```bash
 tail -f ~/.claude/yolt.log
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -308,10 +308,26 @@ def aggregate_decisions(decisions):
 
 def check_unsafe_flags(cmd_args, spec):
     """Check whether cmd_args contains a flag combination the command spec
-    declares unsafe. Returns a human-readable description, or None."""
+    declares unsafe. Returns a human-readable description, or None.
+
+    Recognized fields on `spec`:
+
+    - `unsafe_flags_without_value`: list of flags that are unsafe by their
+      presence alone (`find -delete`).
+    - `unsafe_flag_values`: dict flag -> list of values. Match is
+      case-insensitive equality against the value (`curl -X POST`).
+    - `unsafe_flag_any_value`: list of flags unsafe whenever they appear,
+      regardless of value (`curl -d ...`).
+    - `unsafe_flag_value_prefix`: dict flag -> fnmatch glob pattern.
+      The flag carries a value (inline `--flag=value` or split
+      `--flag value`); the value is matched against the glob
+      (`find -exec rm`, `gh api --input body.json`). `"*"` matches any
+      value.
+    """
     unsafe_flag_values = spec.get("unsafe_flag_values", {})
     unsafe_flag_any_value = set(spec.get("unsafe_flag_any_value", []))
     unsafe_flags_without_value = set(spec.get("unsafe_flags_without_value", []))
+    unsafe_flag_value_prefix = spec.get("unsafe_flag_value_prefix", {})
 
     i = 0
     while i < len(cmd_args):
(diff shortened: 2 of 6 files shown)


debedb (commented):
Two blocking issues before this should merge:

1. `load_shell_rules()` never calls `validate_shell_rules()`, so the normal runtime paths in `grammar_classifier.py` and `yolt_analyzer.py` still accept drift silently, including `~/.claude/yolt/shell.json` overrides. Right now the new validation only runs in unit tests that call it directly. Please wire validation into the actual load path after merging overrides so malformed rule data fails at load time, not only in tests.

2. `validate_shell_rules()` validates nested interpreter module keys but not nested-module `default` values. `_classify_nested_module()` only understands `"safe"` and `"unsafe"`, so a value like `{"default": "bogus"}` currently passes validation and silently degrades to `unknown` at runtime. Please validate `interpreters.*.nested_modules.*.default` explicitly and add a regression test for that case.

The `unsafe_flag_value_prefix` behavior itself looks good to me; the stop here is the validation path not fully enforcing what the PR says it enforces.

debedb:
Both addressed in 08fba1f:

1. **Validation wired into load path.** `load_shell_rules()` now calls `validate_shell_rules()` after the override merge and raises a new `ShellRulesValidationError` on drift. The CLI in `grammar_classifier.py` propagates the error to the terminal; the hook entry in `yolt_analyzer.py` catches it, logs `decision="rules-validation-error"` with the offending keys in `reason`, and exits silently so the session is not broken. A `validate=False` kwarg is provided for tests that need to construct intentionally-malformed fixtures. New `TestLoadShellRulesValidation` covers: real rules load clean through the validating path, malformed default raises, malformed override raises, `validate=False` bypasses.

2. **Nested module `default` validated.** `validate_shell_rules()` now checks `interpreters.*.nested_modules.*.default` against `{"safe", "unsafe"}` — the two values `_classify_nested_module()` actually resolves. Anything else is flagged. Three new cases: `bogus` flagged, `safe` accepted, `unsafe` accepted.

End-to-end coverage in new `TestHookMalformedShellOverride` plants a malformed `~/.claude/yolt/shell.json` under a fake `HOME`, runs the hook, and asserts both the silent-fallthrough response and the `rules-validation-error` log entry — covers both the unknown-key case and the bad-nested-module-default case.

README updated to document the new log decision value.

246 tests pass.

PR #20, opened 2026-05-12
Resolve imported aliases before matching destructive Python calls

### LINKED ISSUE #16: Resolve imported aliases before matching destructive Python calls
Follow-up to #14.

## Problem

The Python analyzer matches destructive calls by the call expression name, but it does not currently resolve imported aliases or direct imports before matching. That makes it easy to hide dangerous calls behind normal Python import forms.

## Reproductions

```bash
python3 -c 'import os as x; x.system("rm -rf /tmp/x")'
python3 -c 'from os import system; system("rm -rf /tmp/x")'
```

Both currently classify as `safe`.

## Root cause

The analyzer records imported modules, but call matching is still done against the surface AST name:

- `x.system(...)` stays `x.system`
- `system(...)` stays `system`

Neither resolves back to `os.system`, so destructive patterns are missed.

## Design

Build a small symbol table during AST traversal for import bindings and normalize call targets before matching rules.

### Supported forms for this release

- `import mod`
- `import mod as alias`
- `from mod import name`
- `from mod import name as alias`

### Out of scope for this release

- full rebinding analysis
- interprocedural data flow
- alias tracking through assignments
- proving object identity for arbitrary locals

## Suggested implementation

1. Record import bindings while visiting `Import` and `ImportFrom`.
2. When analyzing a call target:
   - resolve leading names through the import table
   - rebuild the normalized dotted path when the binding is clear
3. Run existing rule matching against the normalized name.
4. Preserve current conservative behavior for unresolved names:
   - if a name cannot be resolved, do not guess
   - only classify as destructive when the normalized target clearly matches a destructive pattern

## Examples of desired normalization

- `import os as x; x.system(...)` -> `os.system`
- `from os import system; system(...)` -> `os.system`
- `from shutil import rmtree as wipe; wipe(...)` -> `shutil.rmtree`

## Tests

Add regression coverage for:

- alias import of `os.system`
- direct import of `os.system`
- alias import of `shutil.rmtree`
- safe direct imports that should remain safe
- explicit unresolved / rebound cases to document the current boundary

## Acceptance criteria

- The reproductions above classify as `unsafe`.
- Existing direct-call behavior stays intact.
- The implementation documents the supported import-resolution boundary clearly.

debedb:
## Summary

- Add import-binding resolution in `hooks/yolt_analyzer.py`. `SafetyAnalyzer` builds an alias table during `visit_Import` / `visit_ImportFrom`; `_get_call_name` rewrites the leading segment of each call target via that table before pattern matching.
- Supported forms: `import mod`, `import mod as alias`, `import mod.sub`, `import mod.sub as alias`, `from mod import name`, `from mod import name as alias`. Anything the analyzer cannot resolve statically (variable rebinding, `from mod import *`, relative imports) is left at its surface name rather than guessed.
- Existing boto3-style `var.delete_item()` fallback is preserved — the `_matches_pattern` method-name fallback still handles object-identity cases the alias table cannot reach.

After this change:

```
import os as x; x.system("rm -rf /tmp/x")            -> unsafe
from os import system; system("rm -rf /tmp/x")       -> unsafe
from shutil import rmtree as wipe; wipe("/tmp/x")    -> unsafe
```

Closes #16.

## Test plan

- [x] `python3 -m unittest discover -v tests` — 268 tests pass (+22 new)
- [x] New `tests/test_yolt_analyzer.py` covers:
  - [x] Three issue repros
  - [x] Direct-call regression: `import os; os.system(...)`, `import shutil; shutil.rmtree(...)`
  - [x] Safe baselines: `json.dumps`, aliased `json as j`, `from json import dumps`, `import os.path as p; p.exists(...)`, `from requests import get`
  - [x] Boundary cases: `from os import *`, variable rebinding `f = os.system; f(...)`, unimported `os.system(...)`, relative `from . import x`
  - [x] Alias-table state assertions for all five supported binding forms
- [x] New e2e cases in `tests/test_grammar_classifier.py` route the three issue repros through `python3 -c`
- [x] README's Python-rules section documents the supported import forms and out-of-scope cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,23 @@ Rules use `trigger_imports` to scope checks. For example, boto3 patterns
 only apply when `boto3` is imported, so `cache.delete_item()` in a
 non-AWS script doesn't false-positive.
 
+The analyzer also resolves import bindings before matching, so the rule
+patterns (`os.system`, `shutil.rmtree`, ...) catch the same call written
+through any of the standard import forms:
+
+- `import mod`
+- `import mod as alias`
+- `import mod.sub` / `import mod.sub as alias`
+- `from mod import name`
+- `from mod import name as alias`
+
+For example, `from os import system; system("rm -rf /tmp/x")` and
+`import os as x; x.system(...)` both normalize to `os.system` and
+classify as destructive. Out of scope for this release: variable
+rebinding (`f = os.system; f(...)`), `from mod import *`, and
+relative imports — anything the analyzer cannot resolve statically
+is left at its surface name rather than guessed.
+
 ## Custom rules
 
 ### Python rules - `~/.claude/yolt/rules.json`
--- a/README.md
+++ b/README.md
@@ -364,6 +364,22 @@ including assignments inside top-level `if`/`for` blocks) drops the
 binding; function-internal rebinds keep the module-level binding
 intact since they have their own scope.
 
+Module-scope calls resolve against the binding snapshot effective at
+their source line, so a call that appears *before* a later rebind /
+re-import still sees its original binding. For example:
+
+```python
+from os import system
+system("rm -rf /tmp/x")          # unsafe (resolves to os.system)
+system = print                   # later rebind does not retroactively
+                                 # un-flag the earlier call
+```
+
+Calls inside function or class bodies resolve against the final module
+snapshot (after all top-level events), since the body executes when
+the enclosing function/class is invoked rather than at module-load
+time.
+
 Still out of scope: variable rebinding via attribute access,
 `from mod import *`, and relative imports (`from . import x`).
 Anything the analyzer cannot resolve statically is left at its surface
--- a/README.md
+++ b/README.md
@@ -348,10 +348,26 @@ through any of the standard import forms:
 
 For example, `from os import system; system("rm -rf /tmp/x")` and
 `import os as x; x.system(...)` both normalize to `os.system` and
-classify as destructive. Out of scope for this release: variable
-rebinding (`f = os.system; f(...)`), `from mod import *`, and
-relative imports — anything the analyzer cannot resolve statically
-is left at its surface name rather than guessed.
+classify as destructive.
+
+Bindings are collected in a pre-pass over the parsed module body before
+the call walk, so traversal order does not matter — a call inside a
+function defined *before* the matching import still resolves through
+the binding.
+
+Only top-of-file unconditional imports are honored. Imports nested
+under control flow (`if cond: import x`, dead `if False:` branches,
+`try`/`except`, `with`, or inside a function/class body) are NOT
+applied — we cannot statically prove they execute. Top-level
+reassignment of a bound name (`from os import system; system = print`,
+including assignments inside top-level `if`/`for` blocks) drops the
+binding; function-internal rebinds keep the module-level binding
+intact since they have their own scope.
+
+Still out of scope: variable rebinding via attribute access,
+`from mod import *`, and relative imports (`from . import x`).
+Anything the analyzer cannot resolve statically is left at its surface
+name rather than guessed.
 
 ## Custom rules
 
--- a/README.md
+++ b/README.md
@@ -375,10 +375,12 @@ system = print                   # later rebind does not retroactively
                                  # un-flag the earlier call
 ```
 
-Calls inside function or class bodies resolve against the final module
-snapshot (after all top-level events), since the body executes when
-the enclosing function/class is invoked rather than at module-load
-time.
+Calls in deferred positions — `def` / `async def` / `lambda` bodies —
+resolve against the *final* module snapshot, since those bodies execute
+when the function is invoked rather than at module-load time. Calls in
+positions that run at module load — class bodies, decorators, default
+and keyword-default argument values, return annotations — resolve
+against the position-aware snapshot like any other module-scope call.
 
 Still out of scope: variable rebinding via attribute access,
 `from mod import *`, and relative imports (`from . import x`).
--- a/README.md
+++ b/README.md
@@ -378,11 +378,17 @@ system = print                   # later rebind does not retroactively
 Calls in deferred positions — `def` / `async def` / `lambda` bodies —
 resolve against the *final* module snapshot, since those bodies execute
 when the function is invoked rather than at module-load time. Calls in
-positions that run at module load — class bodies, decorators, default
-and keyword-default argument values, parameter annotations (positional,
-positional-only, keyword-only, `*args`, and `**kwargs`), and return
-annotations — resolve against the position-aware snapshot like any
-other module-scope call.
+positions that unconditionally run at module load — class bodies,
+decorators, default and keyword-default argument values — resolve
+against the position-aware snapshot like any other module-scope call.
+
+Annotation expressions (parameter and return) are intentionally not
+analyzed. Under `from __future__ import annotations` (PEP 563) the
+annotation is stored as a string at runtime and never evaluated; PEP
+649 makes lazy annotation evaluation the default in newer Python.
+Flagging annotations would create false positives for modules that
+opted into deferred annotations, and a destructive call hidden inside
+a type hint is not a credible attack pattern.
 
 Still out of scope: variable rebinding via attribute access,
 `from mod import *`, and relative imports (`from . import x`).
--- a/README.md
+++ b/README.md
@@ -379,8 +379,10 @@ Calls in deferred positions — `def` / `async def` / `lambda` bodies —
 resolve against the *final* module snapshot, since those bodies execute
 when the function is invoked rather than at module-load time. Calls in
 positions that run at module load — class bodies, decorators, default
-and keyword-default argument values, return annotations — resolve
-against the position-aware snapshot like any other module-scope call.
+and keyword-default argument values, parameter annotations (positional,
+positional-only, keyword-only, `*args`, and `**kwargs`), and return
+annotations — resolve against the position-aware snapshot like any
+other module-scope call.
 
 Still out of scope: variable rebinding via attribute access,
 `from mod import *`, and relative imports (`from . import x`).
--- a/hooks/yolt_analyzer.py
+++ b/hooks/yolt_analyzer.py
@@ -38,7 +38,27 @@ def load_rules(rules_dir, user_overrides_path=None):
 
 
 class SafetyAnalyzer(ast.NodeVisitor):
-    """AST visitor that checks Python code against safety rules."""
+    """AST visitor that checks Python code against safety rules.
+
+    Import-binding resolution
+    -------------------------
+    During traversal we record the local name introduced by each import so
+    that calls written via aliases or `from`-imports can be matched against
+    rule patterns that use the original dotted path. Supported forms:
+
+      - `import mod`                       -> {"mod": "mod"}
+      - `import mod as alias`              -> {"alias": "mod"}
+      - `import mod.sub`                   -> {"mod": "mod"}
+      - `import mod.sub as alias`          -> {"alias": "mod.sub"}
+      - `from mod import name`             -> {"name": "mod.name"}
+      - `from mod import name as alias`    -> {"alias": "mod.name"}
+
+    `_resolve` rewrites the leading binding of each call target via this
+    table. Anything not bound by one of those import forms is left
+    unchanged - we never guess. Out of scope for this release: variable
+    rebinding (`f = os.system; f(...)`), reassignment through attribute
+    access, conditional / starred imports, and proving object identity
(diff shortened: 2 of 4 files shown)


debedb (commented):
Two correctness issues still block this for me:

1. The alias resolution is source-order-dependent, so supported import forms still miss destructive calls when the binding appears later in the module. `visit_Import` / `visit_ImportFrom` populate `alias_table` during the same traversal that checks calls, and `NodeVisitor` descends into earlier function/class bodies before reaching later top-level imports. For example, both of these currently classify safe even though they should be caught under the supported import forms:

```python
def f():
    system("rm -rf /tmp/x")
from os import system
f()
```

```python
def f():
    x.system("rm -rf /tmp/x")
import os as x
f()
```

This needs either a pre-pass to collect import bindings before call matching, or some equivalent scope-aware restructuring.

2. The implementation still guesses through some out-of-scope bindings instead of leaving names unchanged. Because `alias_table` is global and never invalidated, imports inside dead/conditional blocks and rebinding of imported names both leak through. For example these are currently flagged as `os.system`:

```python
if False:
    from os import system
system("hello")
```

and

```python
from os import system
system = print
system("hello")
```

That contradicts the documented boundary that unresolved / out-of-scope cases should stay at their surface name rather than being guessed.

The straightforward top-level repros from #16 look good; the remaining problem is that the current alias table is not scope/order-safe yet.

debedb:
Both addressed in 0acf118.

**1. Order independence via pre-pass.** `SafetyAnalyzer.analyze()` now calls `_collect_top_level_bindings()` before `self.visit(tree)`. The pre-pass walks only the module-level body, so calls inside a function defined earlier in source than its matching import resolve correctly. Both of your repros now classify as unsafe:

```python
def f():
    system("rm -rf /tmp/x")
from os import system
f()                                  # -> unsafe (os.system)

def f():
    x.system("rm -rf /tmp/x")
import os as x
f()                                  # -> unsafe (os.system)
```

**2. Scope-safe bindings.** The pre-pass only honors imports at the module top level — imports nested under `if`/`while`/`for`/`try`/`with` or inside function/class bodies are skipped entirely. Top-level reassignment of a bound name drops the binding (we cannot prove which definition wins). Function-internal rebinds keep the module-level binding intact since they introduce a separate scope.

```python
if False:
    from os import system
system("hello")                      # -> safe (was unsafe)

from os import system
system = print
system("hello")                      # -> safe (was unsafe)

from os import system
def f():
    system = print
system("rm -rf /tmp/x")              # -> unsafe (function-internal rebind, unchanged)
```

New `TestImportScopeAndOrder` covers all four reviewer cases plus: `if`-guarded conditional imports, `try`/`except` imports, function-internal imports, `for`-target rebinds, top-level `if`-guarded rebinds, and the function-local rebind invariant. The original `TestImportAliasResolution` cases still pass unchanged. 278 tests total.

README updated to spell out the scope rules and the pre-pass design.

debedb (commented):
One more correctness hole still blocks this for me.

The new pre-pass makes alias/rebind state effectively global for the whole module, so a later top-level binding change can rewrite or invalidate an earlier call. That means earlier destructive calls can now disappear incorrectly.

For example, both of these currently classify safe:

```python
from os import system
system("rm -rf /tmp/x")
system = print
```

```python
from os import system
system("rm -rf /tmp/x")
from pathlib import Path as system
```

and likewise:

```python
import os as x
x.system("rm -rf /tmp/x")
x = print
```

The destructive call happens before the later rebinding or re-import, so it should still resolve as `os.system`. Right now `_collect_top_level_bindings()` builds the final binding table first and then uses that same table for the entire call walk, which loses execution-order information at module scope.

So this fixes the two earlier review items, but it still needs module-scope order sensitivity for top-level calls before I’d consider it mergeable.

debedb:
Addressed in 6f23a45.

The pre-pass now builds a line-ordered `_module_events` list (each import contributes a bind event at its line; each module-scope rebind contributes a drop event at its line). For each module-scope call, `_resolve` replays only the events with `lineno < call.lineno` and resolves the leading name against that position-specific snapshot. A `_scope_depth` counter (incremented in overridden `visit_FunctionDef` / `visit_AsyncFunctionDef` / `visit_ClassDef`) routes function-body calls to the final snapshot, since those bodies execute when the enclosing function is invoked rather than at module-load time.

All three of your repros now classify as `unsafe`:

```python
from os import system
system("rm -rf /tmp/x")          # unsafe (was safe)
system = print

from os import system
system("rm -rf /tmp/x")          # unsafe (was safe)
from pathlib import Path as system

import os as x
x.system("rm -rf /tmp/x")        # unsafe (was safe)
x = print
```

Existing behavior preserved:

- `from os import system; system = print; system("hello")` -> safe (call is after the drop event)
- Function-internal calls resolve against the final snapshot (review item 1 fix unchanged)
- Function-local rebinds do not invalidate module-level bindings

New tests cover the three reviewer examples plus a mixed case where two imports straddle two calls (the earlier call resolves to the earlier binding, the later call to the later binding), and an explicit assertion that function-body calls use the final snapshot.

278 prior tests still pass; 283 total. README updated to spell out the position-aware module-scope rule.

debedb (commented):
One more correctness issue still blocks this.

The new line-aware module-scope handling fixes the plain top-level call cases, but the implementation now treats every call visited under `FunctionDef` / `AsyncFunctionDef` / `ClassDef` as if it executes against the final module snapshot. That is not true for all of those AST positions in Python.

In particular, class bodies execute during module loading, and function-definition-time expressions like decorators and default arguments also execute while the module is being defined. Those should therefore resolve against the binding state effective at that point in module execution, not the final snapshot after later top-level rebinds.

For example, both of these currently classify safe:

```python
from os import system
class C:
    system("rm -rf /tmp/x")
system = print
```

```python
from os import system
def f(x=system("rm -rf /tmp/x")):
    return x
system = print
```

In both cases the destructive call happens before the later top-level rebind, so it should still resolve as `os.system`.

The root cause is the `_scope_depth` shortcut: `visit_FunctionDef` / `visit_AsyncFunctionDef` / `visit_ClassDef` increment depth for the entire traversal, and `_resolve()` uses the final snapshot whenever `self._scope_depth > 0`. That conflates true runtime function-body execution with definition-time expressions and class-body execution.

So the previous module-scope blocker is fixed, but this is still not execution-order correct yet.

debedb:
Addressed in 2596b1b.

The `_scope_depth` shortcut was too coarse — it bumped depth for the entire `FunctionDef` / `AsyncFunctionDef` / `ClassDef` subtree, treating decorators, default arg values, and class bodies as deferred even though those expressions run at module-load time.

Replaced with `_visit_function_like` that splits the function-like def into definition-time parts (decorators, positional defaults, kw-only defaults, return annotation) visited at the current scope depth, and the body which is the only genuinely deferred part. `_scope_depth` now bumps only across the body. `visit_ClassDef` does a plain `generic_visit` with no depth bump so class bodies, decorators, bases, and keyword args all use the position-aware snapshot. `visit_Lambda` handles lambdas analogously — defaults at module scope, body deferred.

All three of your repros (and a decorator-call case) now classify `unsafe`:

```python
from os import system
class C:
    system("rm -rf /tmp/x")          # unsafe (was safe)
system = print

from os import system
def f(x=system("rm -rf /tmp/x")):    # unsafe (was safe)
    return x
system = print

from os import system
@system("rm -rf /tmp/x")             # also unsafe
def f():
    pass
system = print
```

Function-body deferral preserved — e.g. a method body inside a class still uses the final snapshot:

```python
from os import system
class C:
    def m(self):
        system("hello")              # uses final snapshot -> safe
system = print
```

New tests cover class body, positional default, kw-only default, decorator, async-function default, lambda default (all flagged) plus the method-body-deferred and lambda-body-deferred invariants. 283 prior tests still pass; 291 total. README updated to spell out the deferred-vs-load-time positions.

debedb (commented):
One remaining correctness gap still blocks this for me.

The new definition-time split in `_visit_function_like()` is manually visiting decorators, positional defaults, kw-only defaults, and the return annotation, but it never visits *parameter annotations* at all. That means calls inside argument annotations are now silently skipped even though equivalent calls in return annotations are still analyzed.

For example, both of these currently classify safe:

```python
from os import system
def f(x: system("rm -rf /tmp/x")):
    return x
system = print
```

```python
from os import system
def f(*, x: system("rm -rf /tmp/x")):
    return x
system = print
```

The root cause is that `_visit_function_like()` no longer does a generic walk of `node.args`; it explicitly visits `node.args.defaults`, `node.args.kw_defaults`, and `node.returns`, but never the `annotation` field on positional-only / positional / kw-only args or vararg / kwarg annotations.

So the class-body / default-arg issue is fixed, but this change still leaves a real false-negative path for definition-time expressions.

debedb:
Addressed in 63e102e.

`_visit_function_like` now walks the `.annotation` field on every `ast.arg`, not just `node.returns`. Added a small helper `_all_arg_nodes(arguments)` that yields every `ast.arg` across `posonlyargs`, `args`, `kwonlyargs`, plus the optional `vararg` and `kwarg`. Each annotation is visited at the current scope depth so it resolves against the position-aware snapshot like any other definition-time expression.

Both of your repros now classify `unsafe`, plus the analogous positional-only, `*args`, `**kwargs`, and async cases:

```python
from os import system
def f(x: system("rm -rf /tmp/x")):       # unsafe (was safe)
    return x
system = print

from os import system
def f(*, x: system("rm -rf /tmp/x")):    # unsafe (was safe)
    return x
system = print
```

New tests cover positional, kw-only, positional-only, `*args`, `**kwargs`, and async-function arg annotations. 291 prior tests still pass; 297 total. README updated to enumerate the annotation positions that evaluate at module load.

debedb (commented):
One remaining correctness issue still blocks this for me.

The new parameter-annotation traversal fixes the earlier gap, but it is still modeling annotation expressions as if they execute unconditionally at function/class definition time. In the current Python runtime that is not generally true: annotations are deferred until `__annotations__` is read, and with `from __future__ import annotations` they are stringized instead of executed.

That means this PR is now creating real false positives by treating annotation calls as destructive side effects in code paths that do not actually execute them. For example this currently classifies unsafe:

```python
from os import system
def f(x: system("rm -rf /tmp/x")):
    return x
```

and this also currently classifies unsafe even though the annotation is explicitly postponed into a string:

```python
from __future__ import annotations
from os import system
def f(x: system("rm -rf /tmp/x")):
    return x
```

I verified this directly in the interpreter: the function definition itself does not run the annotation expression, and under `from __future__ import annotations` the stored annotation is just a string. So expanding the analyzer to flag parameter/return annotations as unconditional destructive calls is not execution-model correct.

The earlier alias-resolution blockers are fixed, but this latest annotation handling still looks too eager to merge.

debedb:
Addressed in 45d2c35.

You're right — flagging annotation expressions is execution-model wrong under `from __future__ import annotations` (PEP 563 stringizes them) and under PEP 649 which makes lazy evaluation the default in newer Python. A destructive call hidden inside a type hint is not a credible attack pattern either way.

Removed the annotation walk entirely from `_visit_function_like`. The helper `_all_arg_nodes` is gone. `_visit_function_like` now visits only decorators, positional defaults, and kw-only defaults at the current scope depth, then bumps `_scope_depth` across the body.

Both of your repros now classify `safe`:

```python
from os import system
def f(x: system("rm -rf /tmp/x")):
    return x                        # safe (was unsafe)

from __future__ import annotations
from os import system
def f(x: system("rm -rf /tmp/x")):
    return x                        # safe (was unsafe)
```

All the other definition-time positions are still flagged unchanged: class body, decorator, positional / kw-only defaults, lambda defaults. Function / lambda body deferral preserved.

The six rev5 annotation tests are replaced with four boundary-documentation cases that explicitly assert annotations stay unanalyzed (positional, kw-only, return, and the `from __future__ import annotations` form). 291 prior tests still pass; 295 total. README updated to document why annotations are skipped.

debedb (commented):
One remaining correctness issue still blocks this for me.

The annotation rollback is fine, but function-local shadowing is still being guessed through as if the imported binding remains active inside the function body. In Python, an assignment inside a function shadows that name for the whole function scope; it does **not** keep resolving through the module import.

For example, this currently classifies unsafe even though the call is just to `print`:

```python
from os import system
def f():
    system = print
    system("hello")
```

The analyzer reports that as `os.system`, which is a false positive.

The root cause is that every call in a deferred function body still resolves against the final module snapshot, with no function-scope binding/shadowing pass. The docs currently say function-internal rebinds keep the module-level binding intact, but that is not how Python name resolution works for a local assignment like `system = print` — once the name is assigned in the function, references in that scope no longer mean the imported module binding.

So the latest annotation fix is good, but the alias-resolution feature still is not scope-correct for basic function-local rebinding.

Issue #1, opened 2026-04-20
Allowlist matcher gaps: arbitrary-execution wrappers and compound commands


debedb:
Two gaps in Claude Code's `.claude/settings.json` `permissions.allow` allowlist
matcher that came up while tightening read-only AWS permissions for the
`clickagy/kpv2` repo. Filing here because YOLT is the natural place for a
hook-based mitigation.

## 1. No safe way to allowlist common interpreters / wrappers

The following commands are ones a senior engineer wants to run without
prompts during read-only investigation, but each grants arbitrary code
execution if added with a wildcard, so they are deliberately left out of
allowlists today:

- Interpreters: `python3 *`, `python *`, `bash *`, `sh *`, `node *`,
  `bun *`, `ruby *`, `perl *`
- JVM launchers: `java *`, `java -jar *`, `gradle *`, `./gradlew *`
- Package runners: `npx *`, `bunx *`, `uvx *`, `uv run *`
- HTTP/RPC tools used both ways: `gh api *` (any HTTP method),
  `curl *`, `kubectl exec *`

The result is that legitimate read-only invocations
(`python3 -c "print(1+1)"`, `gh api /repos/.../issues`) prompt every
single time, which trains users to click "always allow" and effectively
defeats the safety boundary.

### What would help

A YOLT-style pre-tool-call hook that classifies the *full* invocation
(not just the leading command) as read-only or mutating and short-circuits
the prompt for read-only cases. For example:

- `gh api /repos/X/Y/issues` (GET, default method) → allow
- `gh api -X POST /repos/X/Y/issues` → prompt
- `python3 -c "print(open('/etc/passwd').read())"` → allow if filesystem
  access is read-only
- `python3 -c "import os; os.system('rm -rf /')"` → prompt

The classification can be conservative — when in doubt, fall through to
the existing prompt. The win is reducing the no-doubt-read-only cases.

## 2. Compound shell commands escape the matcher

Even with a comprehensive AWS read-only allowlist
(`Bash(aws ecs list-*)`, `Bash(aws ecs describe-*)`, etc.), this still
prompts:

```bash
for svc in $(aws ecs list-services --cluster X); do
  aws ecs describe-services --cluster X --services "$svc"
done
```

The matcher sees the outer command as `for` (or whatever wraps it —
`bash -c "..."`, a function call, etc.) and never decomposes the inner
`aws ecs list-services` / `describe-services` calls. Every such
investigation script triggers a prompt despite both inner commands
[...]

PR #22, opened 2026-05-13
Fix Python alias resolution for local shadowing

### LINKED ISSUE #21: Fix Python alias resolution for function-local and class-local shadowing
## Summary

The Python import-alias resolver added in #20 still guesses through local shadowing inside deferred and immediate nested scopes. In Python, assigning to a name inside a function or class body shadows the imported binding for that scope; it should not keep resolving through the module-level import alias.

## Current false positives

```python
from os import system
def f():
    system = print
    system("hello")
```

This is currently classified as `os.system`, but the call is to the function-local `print`.

```python
from os import system
class C:
    system = print
    system("hello")
```

This is also currently classified as `os.system`, but the class-body assignment shadows the imported name before the call.

## Expected behavior

- module-scope alias resolution should keep working for supported top-level import forms
- module-scope call ordering should keep working
- function-local and class-local assignments should shadow imported bindings in that scope
- unresolved local names should stay at their surface name rather than being guessed

## Suggested approach

Introduce real scope-aware shadowing instead of resolving every deferred body against the final module snapshot. A reasonable implementation path is:

- keep the current module-scope event replay for top-level ordering
- add scope-local shadowing for function/lambda bodies
- add immediate-execution binding/shadowing for class bodies
- use Python `symtable` or an equivalent scope analysis pass rather than growing ad hoc AST heuristics

## Acceptance tests

```python
from os import system
def f():
    system = print
    system("hello")
```

Should classify safe.

```python
import os as x
def f():
    x = print
    x("hello")
```

Should classify safe.

```python
from os import system
class C:
    system = print
    system("hello")
```

Should classify safe.

debedb:
## Summary

- add scope-aware local shadow suppression for deferred function/lambda bodies using Python's symbol-table analysis
- add ordered class-body local shadowing so class-local assignments/defs stop resolving through module import aliases
- extend analyzer docs and regressions for function-local, nested-function, and class-body shadowing cases

Closes #21.

## Test plan

- [x] `python3 -m unittest discover -s tests -p 'test_yolt_analyzer.py'`
- [x] direct analyzer repros for function-local shadowing, class-body shadowing, and the original late-import unsafe case

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,8 +361,11 @@ under control flow (`if cond: import x`, dead `if False:` branches,
 applied — we cannot statically prove they execute. Top-level
 reassignment of a bound name (`from os import system; system = print`,
 including assignments inside top-level `if`/`for` blocks) drops the
-binding; function-internal rebinds keep the module-level binding
-intact since they have their own scope.
+binding. Function / lambda-local rebinds shadow the imported binding
+within that deferred scope, but do not mutate the module-level
+snapshot. Class bodies execute immediately in their own local
+namespace, so class-local assignments likewise shadow imported names
+for later direct class-body calls.
 
 Module-scope calls resolve against the binding snapshot effective at
 their source line, so a call that appears *before* a later rebind /
--- a/hooks/yolt_analyzer.py
+++ b/hooks/yolt_analyzer.py
@@ -12,6 +12,8 @@ import datetime
 import json
 import os
 import sys
+import symtable
+from collections import defaultdict, deque
 from fnmatch import fnmatch
 from pathlib import Path
 
@@ -53,10 +55,8 @@ class SafetyAnalyzer(ast.NodeVisitor):
         the script imports, calls, then later rebinds or re-imports the
         same name — the earlier call should still see the original
         binding.
-      - `alias_table`: the *final* snapshot, applied to calls that live
-        inside function / class bodies. Those bodies execute when their
-        enclosing function/class is invoked, conceptually after the
-        module-level code finishes, so they see the final state.
+      - `alias_table`: the *final* module snapshot, applied to deferred
+        function / lambda bodies after local shadowing is checked.
 
     Supported import forms (recorded as binding events):
 
@@ -79,8 +79,12 @@ class SafetyAnalyzer(ast.NodeVisitor):
         inside top-level `if`/`for` blocks) are recorded as "drop"
         events at their source line, so a later module-scope call no
         longer resolves through the old binding.
-      - Function-internal rebinds do NOT participate, because those
-        introduce their own scope.
+      - Function / lambda bodies may shadow imported names locally.
+        That local shadowing suppresses alias rewriting inside the
+        deferred scope, but does not mutate the module-level snapshot.
+      - Class bodies execute immediately and have their own ordered
+        local bindings, so they need a class-local shadowing layer on
+        top of the surrounding module snapshot.
 
     `_resolve` rewrites the leading binding of each call target. Names
     that have no recorded binding effective at the call's position are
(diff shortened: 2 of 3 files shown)


debedb (commented):
Round 1 review. Acceptance tests from #21 (function-local shadow, alias shadow, class-local shadow at module level) pass. Found one blocking false positive and two non-blocking notes.

## Blocking

**1. Class body nested under a function ignores class-local shadowing → false positive.**

`_resolve` only consults `_class_scope_stack` indirectly through `_current_immediate_table`, which is only called when `self._scope_depth == 0`. Inside any deferred scope (function/lambda), it goes straight to `self.alias_table` and skips the class snapshot entirely.

Repros (run against this branch, both classify `unsafe` and report `os.system`):

```python
from os import system
def outer():
    class C:
        system = print
        system("hello")
```

```python
from os import system
def outer():
    class C:
        def system(self): pass
        system("hello")
```

Both should classify safe — the class body is an immediate-execution scope and the local binding shadows the module alias before the call, exactly as in the module-level case the PR already handles.

Issue #21 states "function-local and class-local assignments should shadow imported bindings in that scope" without qualifying where the class lives, so this is in scope for the fix.

Suggested approach: record `scope_depth_at_entry` on the class scope at `visit_ClassDef` push time and let `_resolve` prefer the class snapshot when the topmost class on the stack was entered at the current `_scope_depth` (i.e., we are directly in the class body, not in a nested method). Methods stay deferred and still see `alias_table`, which keeps `test_method_body_still_deferred_inside_class` green.

## Non-blocking

**2. `_scope_key_for_node` collides for multiple lambdas (or same-named nested defs) on one line.** Key is `("function", lineno, name)` and lambdas all share name `"lambda"`. Today this works because `symtable.get_children()` and `ast.iter_child_nodes` both walk in source order, so the `deque` popleft ordering happens to line up. That invariant is undocumented and easy to break. Including `col_offset` in the key would remove the fragility.

**3. Walrus inside a comprehension is dropped.** PEP 572 leaks a walrus binding from a comprehension into the enclosing function scope, but `_walk_immediate_scope` stops at `ListComp`/`SetComp`/`DictComp`/`GeneratorExp`, so `_binding_events_in_immediate_scope` will not see e.g. `[(y := x) for x in xs]; y(...)`. Edge case, but worth a comment if intentional.

## Resolved / not-issues

- README update reads correctly for the module-level case; once #1 is fixed, the "class bodies execute immediately in their own local namespace" claim will hold for nested classes too.
- Lambda-with-shadow-default-arg works (`lambda system=print: system(...)` is safe).
- `_shadow_stack` correctly propagates outer-function shadowing into nested class bodies (`test_class_body_inside_function_sees_outer_shadow`).

debedb:
Addressed in `f8b41ad99115cceddc280a6a56a2a7551e5a5ca2`.

This fixes direct class-body resolution for classes nested under deferred function scopes: direct calls in the class body now resolve against the class-local shadowing snapshot at the class entry depth, while nested methods still stay deferred on the final module snapshot.

I also added regressions for nested class-body assignment shadowing and nested class-body definition shadowing.

I left the two non-blocking notes unchanged in this commit.

debedb (commented):
Round 2. LGTM — approving.

## Resolved

- **R1 blocking (class body nested under deferred scope ignored class-local shadow)** — fixed in `f8b41ad`. The new `_direct_class_body_table` helper plus `scope_depth_at_entry` on the class scope record correctly distinguish "directly in this class body" from "inside a nested method/lambda". Both R1 repros now classify safe. New regression tests `test_nested_class_body_assignment_shadows_import_binding` and `test_nested_class_body_definition_shadows_import_binding` lock the behavior in.

## Verified regressions

Ran the analyzer on this branch against five cases beyond the two new tests:

- method inside class-in-func, no shadow → still flags `os.system` (method body is deferred, doesn't inherit class scope, per Python semantics).
- method inside module-level class, no shadow → still flags.
- method inside class-in-func where class shadows but method body itself does not → still flags (Python doesn't expose class names to methods).
- deep nest (class within class within function), innermost class shadows → safe.
- existing 52 tests + 2 new = 54/54 green.

## Carried forward

- Non-blocking #2 (lambda `_scope_key_for_node` collision risk on same-line lambdas) and #3 (walrus in comprehension leak) remain. Author flagged them as intentionally deferred. Fine to leave for a follow-up issue if/when they actually bite.

Acceptance criteria from #21 all met.

PR #25, opened 2026-05-13
Path-aware CLI classification for nested mutating verbs (#17)

### LINKED ISSUE #17: Make common CLI classification path-aware for nested mutating verbs
Follow-up to #14.

## Problem

Several common CLIs are modeled only at the first subcommand layer. That is too shallow for namespaces where a safe-looking top-level noun contains mutating nested verbs.

Current false-allow examples:

```bash
git tag v1.2.3
docker image rm alpine
kubectl config set-context prod
helm repo add foo https://example.com
```

These currently classify as `safe` because the classifier stops at `tag`, `image`, `config`, and `repo`.

## Architectural question

Do we need to model every flag and argument for `kubectl`, `gh`, `git`, etc.?

Short answer: no for this release, maybe partially as generated metadata later.

The classifier does not need a full vendor-accurate parser for each CLI. It needs enough structure to distinguish obviously read-only command paths from writes, and it needs to fail closed to `unknown` when that structure is ambiguous.

## Proposed strategy

Separate syntax discovery from safety policy.

1. Bash/tree-sitter remains responsible for shell syntax and argv extraction.
2. CLI-specific logic extracts a command path such as:
   - `git tag`
   - `docker image rm`
   - `kubectl config set-context`
   - `gh pr view`
3. Classification happens against a nested rule tree / trie over that command path.
4. Only classification-relevant flags need special handling:
   - flags that change method / verb semantics
   - flags that inject executable or request input
   - flags that consume the next token and therefore affect path extraction
5. If argv interpretation is ambiguous before we know the path, fall back to `unknown`.

## Why not fully model every flag now

- It does not materially improve safety for the next release's known false-allows.
- Safety semantics do not come from completions or docs; those sources mainly describe syntax.
- Exhaustive per-version flag tracking is expensive and brittle.
- A partial but conservative parser is enough if unmapped cases fail closed to `unknown`.

## Possible longer-term extension

We can optionally ingest official completions/manpages to generate parser metadata:

- subcommand trees
- known valueless flags
- known flags that consume a value

But generated metadata should feed argv parsing only. Safety policy should remain explicit rule data, not something inferred from docs.

## Suggested implementation

1. Extend the rule format to support nested command paths beyond the current top-level subcommand.
2. Rework subcommand classification so it can walk deeper while skipping known flag/value pairs conservatively.
3. Start with namespaces already showing false-allows:
   - `git tag`
   - `docker image|container|volume|network|system`
   - `kubectl config`
   - `helm repo`
4. Remove optimistic top-level `safe` treatment for write-capable namespaces that are only partially modeled.

## Tests

Add regression coverage for:

- `git tag v1.2.3`
- `docker image rm alpine`
- `kubectl config set-context prod`
- `helm repo add foo https://example.com`

Also add representative safe reads in the same namespaces where relevant.

## Acceptance criteria

- The reproduced commands above no longer classify as `safe`.
- The rule format supports at least one nested level beyond the current top-level subcommand.
- Partially modeled namespaces fall back conservatively instead of silently allowing writes.

debedb:
## Summary

- Adds `nested_subcommand` rule data so common CLIs with mutating
  nested verbs (`docker image rm`, `kubectl config set-context`,
  `helm repo add`, `git tag v1.2.3`) no longer silently classify as
  `safe`.
- Introduces a small `empty_decision` field on nested specs so
  `git tag` (bare or `-l`) can keep classifying safe while any
  positional-tagname or mutating flag falls to `unsafe`/`unknown`.
- Removes `image / container / volume / network / system` from the
  flat `docker` safe list (now under `nested_subcommand`); same for
  `kubectl config` and `helm repo`. Bare `docker image`, `kubectl
  config`, `helm repo` now classify `unknown` (ask) instead of
  silently safe.

Closes #17.

## Acceptance

The four reproductions from the issue no longer classify safe:

```
git tag v1.2.3                       -> unknown   (not safe)
docker image rm alpine               -> unsafe
kubectl config set-context prod      -> unsafe
helm repo add foo https://...        -> unsafe
```

Bare reads in the same namespaces stay safe:

```
git tag                              -> safe
git tag -l                           -> safe
docker image ls                      -> safe
kubectl config view                  -> safe
helm repo list                       -> safe
```

## Test plan

- [x] `python3 -m unittest discover -s tests` — all 327 tests pass
      (added ~25 new cases covering acceptance + bare reads +
      partially-modeled-namespace fallback).
- [x] Manual sanity: bare `docker`, `kubectl config`, `helm repo`
      classify `unknown` (conservative fail-closed).
- [x] `git tag -d v1.2.3`, `git tag -a v1 -m "..."`, `git tag -s
      v1` all classify unsafe via `unsafe_flag_any_value`.

### DIFF
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -416,10 +416,13 @@ _ALLOWED_COMMAND_KEYS = frozenset({
     "safe_subcommands", "unsafe_subcommands",
     "safe_subcommand_patterns", "unsafe_subcommand_patterns",
     "nested_subcommand",
+    "empty_decision",
     "service_overrides",
     "safe_operation_patterns", "unsafe_operation_patterns",
     "sql_flags", "sql_file_flags", "sql_positional_index",
 })
+
+_ALLOWED_EMPTY_DECISIONS = frozenset({"safe", "unsafe"})
 
 _ALLOWED_SERVICE_OVERRIDE_KEYS = frozenset({
     "_note",
@@ -519,6 +522,14 @@ def _validate_command_spec(path, spec, errors):
     default = spec.get("default")
     if default is not None and default not in _ALLOWED_DEFAULTS:
         errors.append("{}: unknown default '{}'".format(path, default))
+
+    empty_decision = spec.get("empty_decision")
+    if empty_decision is not None and empty_decision not in _ALLOWED_EMPTY_DECISIONS:
+        errors.append(
+            "{}: empty_decision must be 'safe' or 'unsafe', got '{}'".format(
+                path, empty_decision
+            )
+        )
 
     nested = spec.get("nested_subcommand", {})
     if isinstance(nested, dict):
@@ -795,6 +806,12 @@ class RuleClassifier:
                     return (DECISION_SAFE, "{} {}: read-only".format(cmd_name, sub))
                 if nested_default == "unsafe":
                     return (DECISION_UNSAFE, "{} {}: mutating".format(cmd_name, sub))
+            empty_decision = nested_spec.get("empty_decision")
+            if empty_decision and not rest:
+                if empty_decision == "safe":
+                    return (DECISION_SAFE, "{} {}: read-only (no positional)".format(cmd_name, sub))
+                if empty_decision == "unsafe":
+                    return (DECISION_UNSAFE, "{} {}: mutating (no positional)".format(cmd_name, sub))
             return self._match_subcommand_lists(
                 "{} {}".format(cmd_name, sub), rest, nested_spec
             )
--- a/rules/shell.json
+++ b/rules/shell.json
@@ -457,8 +457,9 @@
       ],
       "nested_subcommand": {
         "repo": {
-          "safe_subcommands": ["list", "ls", "index", "update"],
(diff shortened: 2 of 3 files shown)


debedb (commented):
blocking: `helm.repo` now treats `update` and `index` as read-only via `safe_subcommands`, but both verbs mutate local state. `helm repo update` refreshes the local repo cache, and `helm repo index ./charts` writes or merges an `index.yaml`. With this rule data, those commands still auto-allow as `safe`, which is the same false-allow class this issue is trying to remove.

Concrete repros:

```bash
helm repo update
helm repo index ./charts
```

Both should classify non-safe, and `tests/test_grammar_classifier.py` should not keep `helm repo update` as a safe expectation.

debedb:
Addressed in `dd0eedb`.

`update` and `index` moved from `helm.repo.safe_subcommands` to `helm.repo.unsafe_subcommands`. Test `test_helm_repo_update_safe` flipped to `test_helm_repo_update_unsafe`, and `test_helm_repo_index_unsafe` added for the second repro. All 328 tests pass.

debedb (commented):
LGTM. This round fixes the remaining blocking item: `helm repo update` and `helm repo index` now classify as unsafe, and the tests cover both repros.

Issue #18, opened 2026-05-12
Next release: harden false-allow edges from adversarial review


debedb:
Follow-up to #14.

## Goal

Remove the currently known false-allow classes before expanding coverage further.

This release should focus on correctness of existing architecture rather than adding many new command families.

## Scope

This umbrella covers:

1. Rule-engine correctness for `unsafe_flag_value_prefix`
2. Python import alias resolution in the delegated analyzer
3. Path-aware nested CLI classification for common write-capable namespaces

## Rationale

The broad architecture is already the right one:

- Bash gets a real parser
- Python gets an AST-based delegate
- SQL CLIs get a separate analyzer path
- common CLIs are policy-driven

The immediate problem is not lack of architecture. It is that several current implementations fail open to `safe` in cases where the intended policy is already known.

### 1. Fix rule-engine correctness first

Land support for `unsafe_flag_value_prefix` and add rule-schema validation so policy/config drift cannot silently recur.

### 2. Fix Python delegation second

Add import-binding resolution so the Python analyzer is not trivially bypassed with standard aliasing forms.

### 3. Fix nested CLI semantics third

Move common CLI classification from shallow top-level subcommands to path-aware nested command classification in the namespaces already known to be problematic.

## Command parsing strategy

We do not need to fully model every flag in `kubectl`, `gh`, `git`, or similar CLIs to ship a sound next release.

Recommended approach:

1. Extract a command path from argv.
2. Handle only flags that matter for classification or path extraction.
3. Fall back to `unknown` whenever parsing becomes ambiguous before we have enough information to classify safely.

This keeps the release tractable and conservative.

## Longer-term direction

It is reasonable to explore consuming official bash completions/manpages as parser metadata in a later phase:

- discover subcommand trees
- learn which flags are valueless
- learn which flags consume arguments

But that should remain an input to argv parsing, not the source of truth for safety semantics. Safety policy should remain hand-authored and conservative.

## Release gates

Before shipping, require:

- regression tests for all currently known false-allows
- conservative `unknown` fallback for partially modeled namespaces
- green end-to-end hook tests and classifier tests
- explicit documentation of supported vs unsupported analysis boundaries

## Success criteria

- Known reproductions from #14 no longer classify as `safe`.
- Parser/rule drift is caught in CI.
- The next set of common CLI work can build on a safer nested-classification foundation instead of adding more shallow rules.

Related issues:

- #15 Honor `unsafe_flag_value_prefix` in shell rule evaluation
[...]

PR #30, opened 2026-05-14
Release gates for #18: regression catalogue + CI + analysis-boundaries doc

debedb:
## Summary

Closes three of the release gates from #18 that the sub-issues (#15, #16, #17, #21) did not themselves provide:

1. **Adversarial regression catalogue** (`tests/test_adversarial_regressions.py`) — consolidates the false-allow reproductions closed by PRs #19, #20, #22, #25 into one named suite. Each test cites the issue and PR that fixed it so a future refactor cannot silently re-introduce a known repro.

   - `TestIssue14ShellFalseAllows` — `find -exec / -execdir / -ok / -okdir <cmd>`, `gh api --input <file>` split + inline forms.
   - `TestIssue16PythonImportAliases` — `import os as x; x.system(...)`, `from os import system; system(...)`, `from shutil import rmtree as wipe; wipe(...)`.
   - `TestIssue21PythonLocalShadowing` — function-body and class-body shadow that must not suppress outer unsafe calls.
   - `TestIssue17NestedCliVerbs` — `docker image rm`, `kubectl config set-context`, `helm repo add`, `git tag <name>`, `git tag -d` go `ask`; bare reads (`git tag -l`, `docker image ls`, `kubectl config view`, `helm repo list`) stay `allow`.

2. **CI workflow** (`.github/workflows/ci.yml`) — runs `python -m unittest discover -v tests` on Python 3.10 / 3.11 / 3.12 and runs `validate_shell_rules()` on the bundled `rules/shell.json`. Schema drift becomes a hard CI failure rather than the silent `rules-validation-error` the production hook emits.

3. **`README` "Analysis boundaries" section** — documents in-scope vs out-of-scope for every analysis surface (Bash decomposition, delegated languages, SQL CLIs, Python alias resolution, policy-driven CLIs) and pins the conservative-unknown contract.

## Test plan

- [x] `python3 -m unittest discover -s tests` — 349 tests pass (303 baseline + new regression catalogue).
- [x] `python3 -m unittest tests.test_adversarial_regressions -v` — 21 / 21 green.
- [x] CI dry run locally: `validate_shell_rules(load_shell_rules('rules', validate=False))` returns `[]` → `rules/shell.json: schema OK`.
- [x] README TOC updated to include the new section.

## Out of scope

- Discovered during test authoring that `python3 -c 'import os as x; x.system(...)'` (semicolon one-liner form, no newline between import and call) still slips past alias resolution — multi-line and heredoc forms work, including all forms PR #20 verified. Filing as a separate issue is appropriate; not bundling here to keep the release-gates PR focused.

Refs #18.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,52 @@
+name: CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  test:
+    name: unittest (python ${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Validate bundled rules/shell.json
+        # Hard fail on schema drift. The hook's load-time guard
+        # would already log rules-validation-error and exit
+        # silently in production; CI must fail loudly.
+        run: |
+          python -c "
+          import sys
+          from pathlib import Path
+          sys.path.insert(0, 'hooks')
+          from rule_classifier import load_shell_rules, validate_shell_rules
+          rules = load_shell_rules(Path('rules'), validate=False)
+          errors = validate_shell_rules(rules)
+          if errors:
+              for e in errors:
+                  print('rules/shell.json:', e)
+              sys.exit(1)
+          print('rules/shell.json: schema OK')
+          "
+
+      - name: Run unit + e2e suite
+        run: python -m unittest discover -v tests
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@
 - [Debug / dogfood log](#debug--dogfood-log)
 - [CLI usage](#cli-usage)
 - [Tests and demo](#tests-and-demo)
+- [Analysis boundaries](#analysis-boundaries)
 - [Design principles](#design-principles)
 
 ## Introduction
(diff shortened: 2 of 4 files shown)


debedb (commented):
blocking: this does not actually satisfy issue #18’s release gate as written, because one currently known false-allow is still present on an in-scope surface.

Concrete repro:

```bash
python3 -c 'import os as x; x.system("rm -rf /tmp/x")'
```

The PR body already says that exact `python3 -c` one-liner still slips. But the new regression catalogue in `tests/test_adversarial_regressions.py` avoids that surface by rewriting the #16 alias tests into heredoc form, so CI can go green without covering the still-failing user-visible entry point. Issue #18 is explicitly about removing the *currently known* false-allow classes before release, so this is not just a follow-up nicety; it means the release gate is still open.

Related doc problem: the new README text says `python3 -c` is in scope and even gives `import os as x; x.system(...)` as a destructive example, which overstates current behavior if this PR lands unchanged.

I think this needs either (a) the actual `python3 -c` alias-resolution fix bundled here, with a regression test for the semicolon one-liner form, or (b) the PR scope / README / issue linkage narrowed so it no longer claims the release gate is closed.

debedb:
Addressed in 1f5c1e4 — took path (a).

Root cause: `_snapshot_at_line` keyed binding events by `lineno` only and compared with `>=`. For `import os as x; x.system(...)` the binding event and the call were both at `lineno=1`, so the snapshot at the call's line excluded the binding, alias resolution failed, and `x.system(...)` did not normalize to `os.system`.

Fix: key events by `(lineno, col_offset)` and propagate that tuple through `_snapshot_at_pos`, `_class_snapshot_at_pos`, `_current_immediate_table`, `_direct_class_body_table`, `_table_for_class_definition`, and `_resolve`. Strictly-before-position semantics are preserved, but a binding earlier on the same line is now visible to a later call on that line. Multi-line cases are unchanged because col_offset is irrelevant when the call is on a later line.

Regressions in `tests/test_adversarial_regressions.py::TestIssue16PythonImportAliases`:

```python
def test_import_alias_os_system_dash_c_semicolon(self):
    self.assertEqual(
        self._decision_dash_c(
            'import os as x; x.system(\"rm -rf /tmp/x\")'
        ),
        "ask",
    )
# + from-import unaliased and from-import-as semicolon variants
```

Heredoc variants are kept as `_heredoc` counterparts so both user-visible entry points stay covered.

352 tests pass.

debedb (commented):
LGTM. This round closes the remaining blocking item by fixing same-line `python3 -c` alias resolution and adding direct semicolon one-liner regressions alongside the heredoc cases.

PR #33, opened 2026-05-14
Close lambda key collision + module-walrus drop (issue #24)

### LINKED ISSUE #24: Python analyzer: lambda-key collision and walrus-in-comprehension scope leak (PR #22 follow-up)
Tracks two latent gaps in the Python scope-aware import-alias resolver added in #20 and refined in #22. Both are non-blocking — current tests do not exercise them — but they are real correctness gaps worth fixing before they bite.

## 1. `_scope_key_for_node` collides for multiple lambdas on the same line

`hooks/yolt_analyzer.py::_scope_key_for_node` keys deferred scopes by `("function", node.lineno, "lambda" | node.name)`. Every lambda on a given line shares the key `("function", L, "lambda")`. Shadow sets for the colliding lambdas are appended to the same `deque` and consumed via `popleft` during the AST walk.

Today this happens to produce correct results because `symtable.get_children()` and `ast.iter_child_nodes` both walk in source order, so build order and consumption order line up. The invariant is undocumented and brittle: any future visitor reordering, or symtable behavior change, breaks it silently.

Suggested fix: include `col_offset` in the key (e.g. `("function", lineno, col_offset, name)`) and key the symtable side identically.

Repro that exercises the path today (currently passes — the test is to lock in correctness once the key is made unique):

```python
from os import system
a, b = lambda system=print: system('x'), lambda: system('y')
```

Expected: lambda 1 safe (param shadows), lambda 2 flags `os.system`.

## 2. Walrus inside a comprehension is not recorded in the enclosing immediate scope

PEP 572: an assignment expression (`:=`) inside a comprehension's clauses **leaks** the bound name into the enclosing function scope. `hooks/yolt_analyzer.py::_walk_immediate_scope` stops descent at `ast.ListComp`, `SetComp`, `DictComp`, `GeneratorExp`, so any name bound by a walrus inside such a comprehension is never added to the immediate-scope binding events. The subsequent call to that name therefore still resolves through the module alias table.

Concrete repro (would currently classify unsafe even though `system` was rebound by walrus):

```python
from os import system
def f(xs):
    [s for s in xs if (system := print)]
    system("hello")
```

Suggested fix: detect `ast.NamedExpr` targets inside comprehension descent specifically and record the bound name as an immediate-scope event at the comprehension's lineno.

## Why these are not blockers

- #1 only fires when multiple lambdas share a line *and* one of them shadows an alias that the other does not. Not currently observed in the corpus yolt analyzes.
- #2 only fires when production code uses walrus inside a comprehension to rebind a previously-imported name, and then calls that name after the comprehension. Rare in shell-extracted Python.

Both are correctness defects, not security defects: in #1 the wrong shadow set could be applied; in #2 a legitimate local rebind is missed and a safe call may be misclassified unsafe. No false negatives that would let an unsafe call through. Still worth closing.

## References

- Surfaced in https://github.com/voitta-ai/voitta-yolt/pull/22#pullrequestreview-4277468563
- Acknowledged by author in https://github.com/voitta-ai/voitta-yolt/pull/22#issuecomment-4436332545

debedb:
## Summary

Closes #24.

Two correctness gaps in the scope-aware import-alias resolver from #20 / #22:

1. **Lambda key collision on the same line.** `_scope_key_for_node` keyed every lambda as `("function", lineno, "lambda")`, so siblings on the same source line collided into one deque in `_scope_shadow_queues` and `popleft` order was the only thing keeping them apart. The invariant was undocumented and brittle.

2. **Walrus at module scope did not drop the import binding.** `_collect_top_level_bindings` had no branch for `ast.NamedExpr`. A walrus at module scope (or inside a module-level list / set / dict / generator comprehension, per PEP 572) therefore did not drop a previous import binding, and a subsequent module-scope call to the rebound name was still rewritten through the alias.

## Fix 1 — unique deferred-scope keys

`_scope_key_for_node` now includes `col_offset`:

```python
("function", node.lineno, node.col_offset, "lambda" | node.name)
```

`symtable` does not expose `col_offset`, so the build phase pairs symtable entries with AST nodes by `(lineno, name)` group: within each group symtable yields registrations in CPython AST-visit order, which matches the AST nodes sorted by `col_offset` for siblings parsed from the same source span (the colliding case). A length mismatch within a group raises rather than silently mis-shadowing.

Pairing by group rather than by parallel walk also sidesteps the cross-Python-version genexp / listcomp ordering hazard. Comprehension scopes (`genexpr`, and `listcomp` / `setcomp` / `dictcomp` on 3.10 / 3.11) are skipped by name on the symtable side — the analyzer never pushes a deferred-function scope for them — and their inner functions / lambdas are still collected by recursing into the child table.

## Fix 2 — module-scope walrus drop

Add an `ast.NamedExpr` branch to `_collect_top_level_bindings` that emits a drop event at the walrus's source position:

```python
elif isinstance(node, ast.NamedExpr):
    for name in self._names_in_target(node.target):
        events.append((node_pos, name, None))
```

`_walk_module_scope` already descends through comprehensions, so this covers walrus both outside and inside a module-level comprehension. Class-body walrus inside a comprehension is a Python `SyntaxError`, so no equivalent change is needed for `_binding_events_in_immediate_scope` — that path already handles `NamedExpr` outside comprehensions in class bodies.

## Tests

`tests/test_issue_24_scope_followups.py` adds 7 regressions:

- Same-line lambda shadowing — shadowing param on lambda 1, on lambda 2, and a three-lambda variant.
- Module-scope walrus outside any comprehension.
- Module-scope walrus inside a list comprehension.
- Module-scope walrus inside a generator expression.
- Position-aware drop: a call BEFORE the walrus still resolves through the import; a call AFTER does not.

## Test plan

- [x] `python3 -m unittest discover -s tests` — 359 tests pass (352 baseline + 7 new).
- [x] Repro 1 (lambda collision): `a, b = lambda system=print: system('x'), lambda: system('y')` — only the second lambda is flagged as `os.system`.
- [x] Repro 2 (walrus): `from os import system; if (system := print): pass; system('hello')` — classified safe (was unsafe).
- [x] Existing scope-aware tests untouched.

### DIFF
--- a/hooks/yolt_analyzer.py
+++ b/hooks/yolt_analyzer.py
@@ -13,7 +13,6 @@ import json
 import os
 import sys
 import symtable
-from collections import defaultdict, deque
 from fnmatch import fnmatch
 from pathlib import Path
 
@@ -116,7 +115,8 @@ class SafetyAnalyzer(ast.NodeVisitor):
         # entry is the set of names that shadow outer bindings.
         self._shadow_stack = []
         # Symbol-table-derived shadow sets keyed by AST scope location.
-        self._scope_shadow_queues = {}
+        # Key shape: ("function", lineno, col_offset, name_or_"lambda").
+        self._scope_shadow_map = {}
         # Immediate class-body scopes layered on top of module scope.
         self._class_scope_stack = []
 
@@ -228,16 +228,21 @@ class SafetyAnalyzer(ast.NodeVisitor):
         lambda scope. If matching metadata is absent, fall back to an
         empty set rather than guessing."""
         key = self._scope_key_for_node(node)
-        queue = self._scope_shadow_queues.get(key)
-        if not queue:
+        shadow = self._scope_shadow_map.get(key)
+        if shadow is None:
             return set()
-        return set(queue.popleft())
+        return set(shadow)
 
     @staticmethod
     def _scope_key_for_node(node):
+        """Build a key that uniquely identifies a deferred function or
+        lambda scope. `col_offset` is required to disambiguate multiple
+        scopes that share a line and a name — e.g. two lambdas on the
+        same source line both have name "lambda" and the same lineno."""
+        col = getattr(node, "col_offset", 0)
         if isinstance(node, ast.Lambda):
-            return ("function", node.lineno, "lambda")
-        return ("function", node.lineno, node.name)
+            return ("function", node.lineno, col, "lambda")
+        return ("function", node.lineno, col, node.name)
 
     def _collect_top_level_bindings(self, tree):
         """Walk the module-level body (only) and build the source-order
@@ -314,6 +319,15 @@ class SafetyAnalyzer(ast.NodeVisitor):
                                     item.optional_vars
                                 ):
                                     events.append((node_pos, name, None))
+                    elif isinstance(node, ast.NamedExpr):
+                        # PEP 572: walrus `name := expr` binds at module
+                        # scope when written there directly or inside a
+                        # comprehension / generator expression that has
+                        # no enclosing function. Drop the previous
+                        # import binding so a later module-scope call to
+                        # `name` is not still rewritten through it.
+                        for name in self._names_in_target(node.target):
+                            events.append((node_pos, name, None))
 
         # Stable sort by (lineno, col_offset) preserves source order.
         events.sort(key=lambda e: e[0])
@@ -652,7 +666,7 @@ class SafetyAnalyzer(ast.NodeVisitor):
             return retval
 
         self.source_lines = source.splitlines()
-        self._scope_shadow_queues = self._build_scope_shadow_queues(source)
+        self._scope_shadow_map = self._build_scope_shadow_map(source, tree)
         self._collect_top_level_bindings(tree)
         self.visit(tree)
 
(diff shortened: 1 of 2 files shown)


debedb (commented):
[codex] blocking: walrus-in-lambda is now misclassified as a module-scope rebind.

The new `ast.NamedExpr` drop in `_collect_top_level_bindings()` runs on every node returned by `_walk_module_scope()`, and that walker still descends into `ast.Lambda` bodies. That makes a lambda-local walrus erase the module import binding even though the lambda has its own scope.

Concrete repro on current head:

```python
from os import system
x = lambda: (system := print)
system("boom")
```

Expected: line 3 still resolves to `os.system` and is flagged unsafe.
Actual on this branch: it classifies safe. The base commit still flags it correctly.

I think the fix is to keep lambda bodies out of the module-scope walk, or otherwise gate the `NamedExpr` drop so it only fires for walruses that really bind in module scope.

debedb:
Addressed in 1b38c10.

`_walk_module_scope` already stopped at FunctionDef / AsyncFunctionDef / ClassDef but not at Lambda. Before the earlier commit on this branch, descending through Lambda was harmless — lambda bodies are single expressions and the existing Assign / AugAssign / For / With branches never matched. Adding `ast.NamedExpr` broke that invariant. Fix: add `ast.Lambda` to the stop set so a walrus inside a lambda body stays inside the lambda, and a subsequent module-scope call to a previously-imported name still resolves through the import.

Tests now cover the verbatim repro plus a nested-list variant:

```python
from os import system
x = lambda: (system := print)
system('boom')   # still flagged os.system
```

361 tests pass.

debedb (commented):
[codex] blocking: the lambda stop fixes body walruses, but it also hides lambda defaults that execute in module scope.

This new `ast.Lambda` stop in `_walk_module_scope()` is too broad. A walrus in the lambda *body* should stay lambda-local, but a walrus in a lambda *default expression* runs in the enclosing scope at definition time and still needs to emit a module-scope drop.

Concrete repro:

```python
from os import system
f = lambda y=(system := print): y
system("boom")
```

Expected: safe. `system := print` runs while the lambda is being defined, so by line 3 the module binding has already been rebound to `print`.
Actual with the new walker shape: line 3 still resolves to `os.system` because `_walk_module_scope()` now returns as soon as it reaches the `Lambda` node and never sees the `NamedExpr` inside the default value.

So the previous blocking issue is fixed, but this introduces the mirror-image regression for lambda defaults. I think the walker needs to exclude lambda *bodies* without skipping lambda default expressions that execute in module scope.

debedb:
Addressed in 056f881.

The earlier commit's blanket `ast.Lambda` stop also hid the lambda's default expressions, which execute at definition time in the enclosing scope. Replaced the stop with explicit descent: for `Lambda` we walk `args.defaults` and `args.kw_defaults`; for `FunctionDef` / `AsyncFunctionDef` we also walk `decorator_list`; for `ClassDef` we walk decorators, bases, and keyword values. In all four cases the body is still skipped — that runs in the inner scope.

Repro now passes:

```python
from os import system
f = lambda y=(system := print): y
system('boom')     # safe
```

Body-walrus case from the previous round is still flagged as expected. New tests cover both the default-leak case and a mixed lambda that has a walrus in both default and body. 363 tests pass.

debedb (commented):
[codex] blocking: direct top-level `def` / `class` statements still miss definition-time walrus rebinds.

The new `_walk_module_scope()` logic for `FunctionDef` / `AsyncFunctionDef` / `ClassDef` is correct as far as it goes, but `_collect_top_level_bindings()` still has:

```python
elif isinstance(stmt, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
    continue
```

so direct top-level `def` / `class` statements never reach that walker at all. That means walruses in definition-time expressions on those statements still fail to drop the import binding even though they execute in module scope.

Concrete repros that currently classify unsafe on head but should be safe:

```python
from os import system
def f(x=(system := print)):
    pass
system("boom")
```

```python
from os import system
class C((system := type("B", (), {}))):
    pass
system("boom")
```

and likewise for top-level decorators / class keyword values.

At runtime those definition-time expressions do rebind `system` in module scope before the later call. The current analyzer still reports `os.system` because the `continue` in `_collect_top_level_bindings()` short-circuits the statement before `_walk_module_scope()` can descend into the default / decorator / base / keyword expressions.

debedb:
Addressed in 6e20390.

The previous commit wired `_walk_module_scope` to descend into a def / class's definition-time expressions (decorators, defaults, kw_defaults, bases, keyword values) while skipping the body, but `_collect_top_level_bindings` short-circuited those statements with an `elif ... continue` so the walker never ran on them. Removed the `continue`; FunctionDef / AsyncFunctionDef / ClassDef now flow through to the same loop that handles other top-level statements, and any walrus inside a definition-time expression emits a module-scope drop event at its own `(lineno, col_offset)`.

Both verbatim repros now classify safe:

```python
from os import system
def f(x=(system := print)): pass
system('boom')      # safe

from os import system
class C((system := type('B', (), {}))): pass
system('boom')      # safe
```

Tests added: top-level def-default, class-base, and decorator variants. 366 tests pass.

debedb (commented):
[codex] LGTM. This round closes the last blocking hole: direct top-level `def` / `class` definition-time walruses now flow through `_walk_module_scope()` and drop the module import binding correctly.

Verified on head `6e20390` against the previously-blocking repros:

- lambda body walrus still flags `os.system`
- lambda default walrus is safe
- top-level def default / class base / decorator / class keyword-value walruses are safe

Focused validation on this branch: `python3 -m unittest tests.test_issue_24_scope_followups tests.test_yolt_analyzer`.

PR #34, opened 2026-05-16
find write-action flags route through safe_write_targets (closes #27)

### LINKED ISSUE #27: find action flags that write files: -fprint, -fprintf, -fls, -fls0
Split from #23.

## Problem

`find` action flags `-exec`, `-execdir`, `-ok`, `-okdir`, `-delete` are already classified unsafe via `unsafe_flag_value_prefix` / `unsafe_flag_any_value` (rules/shell.json:218). Missing: the four `find` actions that write to a path argument with no shell redirect, so they bypass the `safe_write_targets` check entirely.

| Flag | Effect |
| ---- | ------ |
| `-fprint FILE` | append each matched path to `FILE` |
| `-fprintf FILE FORMAT` | append formatted output to `FILE` |
| `-fls FILE` | `-ls` output to `FILE` |
| `-fls0 FILE` | `-fls` with NUL separators |

## Repro

```
find / -name '*' -fprint /etc/profile
```

Classifies safe today (verify), writes to `/etc/profile`.

## Proposed fix

Add these to `find`'s `unsafe_flag_value_prefix` in rules/shell.json. Alternative: route the path argument through the same safe-write check that redirects use — more permissive (allows `-fprint /tmp/x`) and consistent with the existing model. Probably the better shape.

## Also worth verifying in same change

- `-newerXY`, `-printf` with `%` formats — no write, but `%h`, `%f` etc. can leak.
- GNU find `-files0-from -` — read-list from stdin, no write.
- BSD/macOS `find` does not support most of these flags; behavior is GNU-specific.

debedb:
## Summary

- `find` write-action flags `-fprint FILE`, `-fprintf FILE FORMAT`, `-fls FILE`, `-fls0 FILE` no longer classify `safe` regardless of target.
- New per-command spec field `write_flag_value_targets`: list of flags whose value is a write-target path. The path is checked against the existing top-level `safe_write_targets` allow list — same list `> FILE` redirects use — so writes to `/tmp/*` / `~/.cache/*` / `/dev/null` stay safe, writes to `/etc/profile` etc. flip to `ask`.
- Path expansion and fnmatch semantics mirror `grammar_classifier._target_is_safe_write` so flag-value writes and redirect writes resolve the same way.

## Repro

```
find / -name '*' -fprint /etc/profile     # before: safe   after: ask
find / -fprintf /etc/passwd '%p\n'        # before: safe   after: ask
find / -fls    /var/log/x.log             # before: safe   after: ask
find / -fls0   /tmp/list.bin              # before: safe   after: safe (allowed by /tmp/*)
find / -fprint /tmp/list.txt              # before: safe   after: safe (allowed by /tmp/*)
```

## Test plan

- [x] `python3 -m unittest discover -v tests` — 379 / 379 OK on `feature/issue-27-find-write-flags`.
- [x] New unit coverage in `tests/test_rule_classifier.py::TestCheckUnsafeFlags` (7 cases: outside-allow, in-allow, `-fprintf` first-arg-is-path semantics, no-value, empty / None allow list, `~/` home expansion, inline `--flag=value` form).
- [x] New regression coverage in `tests/test_adversarial_regressions.py::TestIssue27FindWriteFlags` (6 cases covering the four flags through the real `yolt_analyzer.py --hook` path).
- [x] Manual smoke matrix via `grammar_classifier.py` — all 9 cases (4 new flag pairs + bare `find` + `-delete` + `-exec`) classify as expected.

## Out of scope

Issue #27 calls out four more areas worth verifying (`-newerXY`, `-printf` format leaks, GNU `-files0-from -`, BSD/macOS find divergence). None of those write files, so they're left for a follow-up — keeping this PR narrow to the bug actually filed.

Closes #27

### DIFF
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -333,7 +333,30 @@ def aggregate_decisions(decisions):
     return (DECISION_SAFE, "; ".join(safe_reasons) if safe_reasons else "no commands")
 
 
-def check_unsafe_flags(cmd_args, spec):
+def _expand_home(path):
+    home = os.environ.get("HOME")
+    if home and path.startswith("~/"):
+        return home + path[1:]
+    return path
+
+
+def _path_matches_safe_write_targets(target, safe_write_targets):
+    """Return True if `target` matches any of `safe_write_targets` under
+    fnmatch semantics. Mirrors the redirect-target check that
+    `grammar_classifier._target_is_safe_write` runs against `> file`
+    redirects, so flag-value writes (e.g. `find -fprint FILE`) and
+    redirect writes use the same allow list."""
+    if not safe_write_targets:
+        return False
+    expanded = _expand_home(target)
+    for pat in safe_write_targets:
+        pat_expanded = _expand_home(pat)
+        if fnmatch(target, pat) or fnmatch(expanded, pat_expanded):
+            return True
+    return False
+
+
+def check_unsafe_flags(cmd_args, spec, safe_write_targets=None):
     """Check whether cmd_args contains a flag combination the command spec
     declares unsafe. Returns a human-readable description, or None.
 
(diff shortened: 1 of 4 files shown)


debedb (commented):
[codex] LGTM. This round correctly routes `find`'s path-writing flags through `safe_write_targets`, and the new rule plus regression coverage matches issue #27's acceptance criteria.

PR #36, opened 2026-05-17
Honor explicit Bash allow patterns and surface workflow hints

### LINKED ISSUE #35: work-on-pr loop: surface paste-ready allow pattern + reconsider 'git add' classification
## Problem

In `work-on-pr` skill loops (canonical: https://github.com/voitta-ai/skillz/blob/master/skills/work-on-pr/SKILL.md), YOLT hook fires `ask` on every `git add` / `git commit` / `git push` even when the operation targets the agent's own feature branch in a worktree. Each round of PR feedback adds 3+ permission prompts — the loop stalls waiting for human approval on its own work.

Example prompts seen in `~/g/git.voitta/voitta-yolt-wt-27` session:

```
Hook PreToolUse:Bash requires confirmation for this command:
 YOLT: Mutating command detected:
   git add: mutating [settings]
 settings.json to update hooks

 Do you want to proceed?
```

(plus same shape for `git commit -F /tmp/commit-27.txt` and `git push -u origin feature/issue-27-find-write-flags`)

### 1. Hook output isn't actionable

The current `ask` body is just `YOLT: Mutating command detected:\n  <reason>` (see `hooks/yolt_analyzer.py:1017`). The user is left to derive *which* allow pattern would have made this safe. For a loop that fires 3 prompts per PR round, this is the dominant friction.

Proposal: when classifying as unsafe, also emit a paste-ready `Bash(...)` allow pattern derived from the command shape. E.g. for `git -C /Users/x/foo-wt-27 push -u origin feature/issue-27-find-write-flags`, suggest:

```
To skip this prompt in future, add to ~/.claude/settings.json#permissions.allow:
  "Bash(git -C * push -u origin feature/*)"
```

Generic worktree path (`*`), branch globbed by prefix (`feature/*`), flags preserved (`-u`). The user copies the line, no need to read the skill docs to compose it.

The same shape works for:
- `git -C * add:*`
- `git -C * commit:*` (covers both `-m "..."` and `-F /tmp/file`)
- `git -C * push origin feature/*` (no-`-u` variant)
- `git -C * push -u origin feature/*` (first-push variant)

### 2. `git add` classification is overconservative for self-PR workflow

`rules/shell.json` lists `add` in `unsafe_subcommands` (line 322). For `work-on-pr`, every round stages files the agent itself just edited in a worktree — staging is local-only and trivially reversed by `git reset`. The CC docs explicitly warn against `git add -A` / `git add .` for grabbing sensitive files, but `git add <explicit-paths>` is much narrower.

Two paths forward:

a) Move `add` to `safe_subcommands`. Risk: `git add -A` / `git add .` still grabs everything in cwd. The user already documents avoiding those in CLAUDE.md, but rule engine can't enforce.

b) Add `nested_subcommand` entry for `add` like the existing `tag` block, where bare `git add <explicit-path>` is safe, and `-A` / `--all` / `.` is unsafe.

Prefer (b). Same pattern as `tag` at `rules/shell.json:336`.

## Related

- Cross-ref skillz issue: <link added after issue created>
- `work-on-pr` SKILL.md `Auto-approved operations` section already documents the allow patterns the user has to add manually today — see https://github.com/voitta-ai/skillz/blob/master/skills/work-on-pr/SKILL.md#auto-approved-operations-self-pr-workflow.
- Existing issue #32 (decide on `--repo` ownership handling) is a different concern — that's about `gh pr merge --repo otherorg/critical`; this one is about local git operations in agent worktrees.

## Out of scope

- Auto-allow `git push` to `master` / default branch — must stay prompting.
- Auto-allow `git rebase` / `git reset --hard` / `git push --force` — destructive.
- The `Edit` / `Write` tool-level prompts (those are CC harness, not YOLT — they go in the skillz fix).

debedb:
## Summary

- treat `permissions.allow` `Bash(...)` entries as an explicit user override for both `unknown` and `unsafe` command decisions, not just `unknown`
- add paste-ready allow-pattern hints for the common self-PR workflow write shapes that still prompt before a user has any allowlist in place (`git add` / `git commit` / `git push`, `gh pr ...`, `gh issue ...`)
- update the classifier docs and tests to match the new contract

## Why this shape

Issue #35 was updated after filing with the verified root cause: YOLT already loads the user's allow patterns, but `_maybe_allow()` only consults them for `unknown` results. That means documented patterns like `Bash(git -C * add:*)` or `Bash(gh issue create:*)` are ignored for the exact `unsafe` commands the `work-on-pr` loop emits.

This change keeps the default rule tables conservative, but treats an explicit user allowlist as authoritative. The new prompt hint remains useful for first-time users who have not populated that allowlist yet.

## Out of scope

- no change to default `git add` / `git push` safety classification in `rules/shell.json`
- no hardcoded `never_allow_even_if_user_says_so` denylist yet
- no expansion yet for `gh api ...` reply endpoints

## Test plan

- [x] `python3 -m unittest -v tests.test_rule_classifier tests.test_grammar_classifier tests.test_yolt_hook.TestHookEndToEnd tests.test_yolt_hook.TestHookAllowlistDiscovery`
- [ ] `python3 -m unittest tests.test_yolt_hook -v`
  - still hits the pre-existing bootstrap/log harness `import-error` failures in this sandboxed environment; not caused by this patch

Closes #35

### DIFF
--- a/hooks/grammar_classifier.py
+++ b/hooks/grammar_classifier.py
@@ -368,7 +368,7 @@ class GrammarClassifier:
 
         sub = argv[i]
         if sub in {"add", "commit"}:
-            return "Bash({} {}:*)".format(" ".join(prefix), sub)
+            return "Bash({} {}*)".format(" ".join(prefix), sub)
 
         if sub != "push":
             return None
@@ -402,7 +402,7 @@ class GrammarClassifier:
             "issue": {"create", "comment", "edit"},
         }
         if action in allowed.get(namespace, set()):
-            return "Bash(gh {} {}:*)".format(namespace, action)
+            return "Bash(gh {} {}*)".format(namespace, action)
         return None
 
     @staticmethod
--- a/hooks/grammar_classifier.py
+++ b/hooks/grammar_classifier.py
@@ -84,6 +84,31 @@ class GrammarClassifier:
             return (DECISION_SAFE, "no commands")
         return aggregate_decisions(decisions)
 
+    def suggest_allow_pattern(self, command):
+        """Best-effort `Bash(...)` allow hint for a single primary command.
+
+        This is intentionally narrow and only covers the self-PR workflow
+        write shapes we document today. Compound shells with multiple
+        primary commands return `None` rather than guessing."""
+        if not command or not command.strip():
+            return None
+
+        src = command.encode("utf-8")
+        tree = self._parser.parse(src)
+        root = tree.root_node
+        if root.has_error:
+            return None
+
+        commands = []
+        self._collect_primary_command_nodes(root, commands)
+        if len(commands) != 1:
+            return None
+
+        argv = self._argv_from_command(commands[0], src)
+        if not argv:
+            return None
+        return self._suggest_allow_pattern_from_argv(argv)
+
     # --- AST walker ---
 
     def _walk(self, node, src, decisions, _depth):
@@ -186,6 +211,21 @@ class GrammarClassifier:
             return
         for c in node.children:
             self._collect_substitutions(c, src, decisions, _depth)
+
+    def _collect_primary_command_nodes(self, node, commands):
+        if node.type in ("command_substitution", "process_substitution",
+                         "function_definition"):
+            return
+        if node.type == "redirected_statement":
+            cmd = self._first_child(node, "command")
+            if cmd is not None:
+                commands.append(cmd)
+                return
+        if node.type == "command":
+            commands.append(node)
+            return
+        for c in node.children:
+            self._collect_primary_command_nodes(c, commands)
 
     # --- Argv reconstruction ---
 
@@ -299,12 +339,71 @@ class GrammarClassifier:
 
     def _maybe_allow(self, match_string, result):
         decision, reason = result
-        if decision != DECISION_UNKNOWN:
+        if decision == DECISION_SAFE:
             return result
         match = match_allow_patterns(match_string, self.allow_patterns)
         if match is None:
             return result
         return (DECISION_SAFE, "matches user allow pattern '{}'".format(match))
+
+    @staticmethod
+    def _suggest_allow_pattern_from_argv(argv):
+        cmd_name = os.path.basename(argv[0])
+        if cmd_name == "git":
+            return GrammarClassifier._suggest_git_allow_pattern(argv)
+        if cmd_name == "gh":
+            return GrammarClassifier._suggest_gh_allow_pattern(argv)
+        return None
+
+    @staticmethod
+    def _suggest_git_allow_pattern(argv):
+        prefix = ["git"]
+        i = 1
+        if i + 1 < len(argv) and argv[i] == "-C":
+            prefix.extend(["-C", "*"])
+            i += 2
+
+        if i >= len(argv) or argv[i].startswith("-"):
+            return None
+
+        sub = argv[i]
+        if sub in {"add", "commit"}:
+            return "Bash({} {}:*)".format(" ".join(prefix), sub)
+
+        if sub != "push":
+            return None
+
+        hint = prefix + ["push"]
+        j = i + 1
+        if j < len(argv) and argv[j] == "-u":
+            hint.append("-u")
+            j += 1
+        if j + 1 >= len(argv):
+            return None
+
+        remote = argv[j]
+        branch = argv[j + 1]
+        if remote != "origin":
+            return None
+
+        branch_pat = "feature/*" if branch.startswith("feature/") else branch
+        hint.extend([remote, branch_pat])
+        return "Bash({})".format(" ".join(hint))
+
+    @staticmethod
+    def _suggest_gh_allow_pattern(argv):
+        if len(argv) < 3:
+            return None
+
+        namespace = argv[1]
+        action = argv[2]
+        allowed = {
+            "pr": {"create", "comment", "edit", "merge", "ready"},
+            "issue": {"create", "comment", "edit"},
+        }
+        if action in allowed.get(namespace, set()):
+            return "Bash(gh {} {}:*)".format(namespace, action)
+        return None
 
     @staticmethod
     def _slice(node, src):
--- a/README.md
+++ b/README.md
@@ -203,9 +203,9 @@ Add to `~/.claude/settings.json`:
 YOLT reads your `permissions.allow` Bash() entries from
 `~/.claude/settings.json`, the project's `.claude/settings.json`, and
 `.claude/settings.local.json`. After the rule classifier runs on each
-AST `command` node, any node that would otherwise be `unknown` is
-upgraded to `safe` if its reconstructed argv matches one of those
-patterns.
+AST `command` node, any node that would otherwise be `unknown` or
+`unsafe` is upgraded to `safe` if its reconstructed argv matches one
+of those patterns.
 
 This earns its keep on compound forms. The built-in matcher only sees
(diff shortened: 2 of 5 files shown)


hooks/grammar_classifier.py
```diff
@@ -299,13 +339,72 @@ def _bash_inline_analyzer(self, script, depth):
...
+        if i + 1 < len(argv) and argv[i] == "-C":
+            prefix.extend(["-C", "*"])
+            i += 2
+
+        if i >= len(argv) or argv[i].startswith("-"):
+            return None
+
+        sub = argv[i]
+        if sub in {"add", "commit"}:
+            return "Bash({} {}:*)".format(" ".join(prefix), sub)
```
debedb: [claude] blocking: the `:*` suffix is Claude Code's outer-matcher convention, not `fnmatch`. YOLT's `match_allow_patterns` uses `fnmatch` directly against `" ".join(argv)`, so `Bash(git -C * add:*)` will not match `git -C /tmp add file` (the `:` vs ` ` mismatches literally). Net effect: the user pastes the suggested hint, CC accepts it, YOLT still prompts — which is the issue this PR set out to fix.

Suggested form: `"Bash({} {}*)".format(" ".join(prefix), sub)` — that matches both fnmatch and CC, and is consistent with the existing `Bash(aws s3 ls*)` style in README.

hooks/grammar_classifier.py
```diff
@@ -299,13 +339,72 @@ def _bash_inline_analyzer(self, script, depth):
...
+            return None
+
+        namespace = argv[1]
+        action = argv[2]
+        allowed = {
+            "pr": {"create", "comment", "edit", "merge", "ready"},
+            "issue": {"create", "comment", "edit"},
+        }
+        if action in allowed.get(namespace, set()):
+            return "Bash(gh {} {}:*)".format(namespace, action)
```
debedb: [claude] blocking: same `:*` vs fnmatch problem as the git hint above. `fnmatch("gh issue create --title x", "gh issue create:*")` is `False`. The hint won't be honored by `_maybe_allow`, so the user will still see the prompt after pasting it.

Suggested form: `"Bash(gh {} {}*)".format(namespace, action)` (or `... {} *)` with a leading space before `*`).

Worth adding a test that round-trips: emit hint for cmd, install hint as the only allow pattern, re-classify cmd, expect `DECISION_SAFE`. That would have caught both findings.

debedb (commented):
[claude] Review of PR #36 (round 1, self-review).

**Blocking — the suggested allow-pattern hints can't be honored by YOLT itself.**

`hooks/grammar_classifier.py` emits hints with a `:*` suffix for `git add` / `git commit` (line 371) and `gh pr|issue <action>` (line 405). YOLT's own `match_allow_patterns` is straight `fnmatch` on the inner pattern (`rule_classifier.py:308`), and `fnmatch` treats `:` as a literal. The reconstructed argv used as the match string is `" ".join(argv)`, which has a space at that position, not a colon.

Verified with the same fnmatch path the hook uses:

```
>>> from fnmatch import fnmatch
>>> fnmatch("git add file.txt", "git add:*")
False
>>> fnmatch("gh issue create --title x", "gh issue create:*")
False
>>> fnmatch("git add file.txt", "git add*")
True
```

Practical effect: a user who follows the new prompt verbatim and pastes `"Bash(gh issue create:*)"` (or the `git -C * add:*` variant) into `~/.claude/settings.json#permissions.allow` will get the line accepted by Claude Code's outer matcher, but YOLT's `_maybe_allow` still returns the original UNSAFE decision because the pattern doesn't match. The hook continues to prompt — which is the exact UX issue #35 set out to fix.

The `git push` hint at line 391 happens to come out clean (`Bash(git -C * push -u origin feature/*)`, no colon) and round-trips correctly, so it masks the bug from the existing tests. The new hint tests assert the hint string is present but never feed the hint back through `_maybe_allow` to confirm it actually upgrades the same command.

Suggested resolution:
1. Change the two `:*` format strings to the trailing-glob form, e.g. `Bash({} {}*)` for git add/commit and `Bash(gh {} {}*)` for the gh family. That matches both fnmatch and CC's outer matcher, and lines up with the existing README examples (`Bash(aws s3 ls*)`).
2. Add one test that asserts: emit a hint for command C, install it as the sole allow pattern, re-classify C, expect SAFE. That would have caught this round.

**Non-blocking observations.**

- `suggest_allow_pattern` re-parses `command` through tree-sitter even though `_classify_command_node` already produced an argv for the same string moments earlier. For the single-primary-command path this is wasted work; the public surface could take argv. Fine as-is for round 1 if you'd rather not entangle the analyzer pipeline.
- The README block in this PR reads cleanly, but it doesn't mention the colon-suffix convention at all — yet the hints emit it. Either drop the colon (preferred, see above) or document that YOLT honors both forms (which would mean teaching `match_allow_patterns` to normalize `pat:*` → `pat *` / `pat*`).

Everything else (the `_maybe_allow` change itself, the new test rewordings, the README updates around explicit-override semantics, the `-C *` / `feature/*` normalization in the git hint) reads correctly and matches the verified root cause in issue #35.

hooks/grammar_classifier.py
debedb: [codex] Changed the `git add` / `git commit` hint form from `:*` to the plain trailing-glob form (`Bash(git -C * add*)` / `Bash(git -C * commit*)`), so YOLT's own `fnmatch` pass now honors the pasted hint instead of only Claude Code's outer matcher. Addressed in 007c5f5.

debedb:
[codex] Addressed the review in 007c5f5.

- switched the hinted allow-pattern forms from `:*` to plain trailing globs so YOLT's `fnmatch` path and Claude Code's outer matcher both accept the same pasted hint
- added a round-trip classifier test for the emitted `git` and `gh` hints so the suggested pattern has to suppress the originating command on a second pass
- reran `python3 -m unittest -v tests.test_rule_classifier tests.test_grammar_classifier tests.test_yolt_hook.TestHookEndToEnd tests.test_yolt_hook.TestHookAllowlistDiscovery`

hooks/grammar_classifier.py
debedb: [codex] Changed the `gh` hint form to `Bash(gh <namespace> <action>*)` and added a round-trip classifier test that emits the hint, installs it as the only allow pattern, and reclassifies the same command to `safe`. That now covers both the `git` and `gh` cases this review called out. Addressed in 007c5f5.

debedb (commented):
[claude] LGTM (self-review, COMMENT in lieu of APPROVE).

Round 2 diff (007c5f5) addresses both blocking findings from round 1:

- **resolved**: `_suggest_git_allow_pattern` now emits `Bash({} {}*)` for `git add` / `git commit`, dropping the `:*` form that fnmatch couldn't honor (grammar_classifier.py:371).
- **resolved**: `_suggest_gh_allow_pattern` likewise emits `Bash(gh {} {}*)` for `gh pr|issue <action>` (grammar_classifier.py:405).
- **resolved**: the suggested round-trip test landed — `test_suggested_allow_hints_round_trip` emits the hint, installs the inner pattern as the sole allow entry, re-classifies the original command, asserts SAFE. That closes the gap that let the colon form slip past round 1.

Spot-checked the new patterns through the same `fnmatch` path `_maybe_allow` uses; all five representative commands (`git add`, `git -C ... add`, `git commit`, `gh issue create`, `gh pr comment`) now match their emitted hints. The `test_gh_issue_create_reason_includes_allow_hint` assertion was also updated to the new string.

Nothing else moved in the diff and no unrelated regressions. PR is ready to merge from this reviewer's perspective.

PR #42, opened 2026-05-26
Add unsafe_write_targets deny list for protected write paths (issue #28)

### LINKED ISSUE #28: Add unsafe_write_targets list for dotfile / config / startup paths
Split from #23.

## Problem

`rules/shell.json#safe_write_targets` whitelists `/tmp/*`, `~/.cache/*`, `~/.claude/*`, etc. Anything not on the list falls through to `unknown`, which prompts. There's no symmetric *deny* list, so a redirect to `~/.bashrc` or `~/.ssh/authorized_keys` is treated the same as a redirect to `./output.txt` — both `unknown`.

The risk profile is asymmetric: writing to `~/.ssh/authorized_keys` is qualitatively different from writing to a project file. It should classify `unsafe` (ask with a specific reason), not `unknown` (default prompt with no context).

## Proposed list

```
~/.bashrc, ~/.zshrc, ~/.profile, ~/.bash_profile, ~/.bash_logout
~/.ssh/authorized_keys, ~/.ssh/config, ~/.ssh/id_*, ~/.ssh/known_hosts
~/.aws/credentials, ~/.aws/config
~/.gitconfig, ~/.git-credentials
~/.npmrc, ~/.pypirc, ~/.netrc
~/Library/LaunchAgents/*
/Library/LaunchDaemons/*
~/.config/systemd/user/*
/etc/systemd/system/*
/etc/cron.*, /var/spool/cron/*
/etc/*
/usr/local/etc/*
```

## Special case: `~/.claude/*`

Currently *safe*-write under `safe_write_targets`. Load-bearing for legitimate cache files. But `~/.claude/settings.json` and `~/.claude/settings.local.json` can neuter the hook itself (add a static `Bash(*)` allow). Carve those out: keep `~/.claude/*` safe-write, but explicitly mark `~/.claude/settings.json` and `~/.claude/settings.local.json` unsafe-write. Same for any plugin manifest path that controls hook registration.

## Application points

This list should apply to:
1. Redirect targets in `redirected_statement` (where `safe_write_targets` is consulted today).
2. Path arguments to `tee`, `cp dst`, `mv dst`, `install -t`, `dd of=`.
3. `find -fprint*` action targets (see sibling issue).
4. `sed -i` already classifies unsafe on the flag alone; no change.

debedb:
## Summary

Adds a top-level `unsafe_write_targets` deny list to `rules/shell.json`
for dotfile / config / startup paths, and routes write targets through
it. A match classifies **`unsafe`** (ask with a specific reason) instead
of **`unknown`** (contextless default prompt).

The deny list is checked **before** `safe_write_targets`, so an entry on
it overrides a broader safe glob.

### Security hole closed

`echo x > ~/.claude/settings.json` previously classified **`allow`**
(auto-run), because `~/.claude/*` is a safe-write target -- even though
`settings.json` can disable this hook. `settings.json` /
`settings.local.json` are now on the deny list and override the safe
glob, so they `ask`. The rest of `~/.claude/*` (e.g. legitimate cache
writes) stays a safe-write target.

## Application points (all four from the issue)

| Point | Where | Behavior change |
| ----- | ----- | --------------- |
| Redirects (`> FILE`, `>> FILE`) | `grammar_classifier._walk_redirected` | deny -> `unsafe`; non-deny non-safe -> `unknown` (unchanged); safe -> benign |
| Flag-value targets (`install -t`, `find -fprint*`) | `check_unsafe_flags` | deny-checked before the existing `safe_write_targets` check |
| Value-prefix targets (`dd of=PATH`) | `check_unsafe_flags` | new `write_value_prefix_targets` |
| Positional targets (`tee` all, `cp`/`mv` last) | `check_unsafe_flags` | new `write_target_all_positional` / `write_target_last_positional` |

`tee`/`cp`/`mv`/`install`/`dd` are already `default: unsafe`, so for them
a deny match upgrades the *reason* to a specific protected-path one; a
non-match falls through to the default. The genuine decision change
(`unknown`/silent -> `unsafe`/ask, and the `settings.json`
`allow` -> `ask`) is on the redirect path.

## Precedence and overrides

- `unsafe_write_targets` is consulted before `safe_write_targets`
  everywhere (the `settings.json` carve-out depends on this).
- Allow patterns still override, consistent with every other unsafe
  atom: an explicit `Bash(echo *)` upgrades a deny-path write to `safe`.

## Tests

- `TestIssue28UnsafeWriteTargets` -- adversarial regression catalogue:
  the `settings.json` false-allow plus redirect / `cp` / `dd` repros and
  benign controls.
- `TestUnsafeWriteTargets` (grammar) and `TestUnsafeWriteTargetArgs`
  (`check_unsafe_flags` unit) coverage, including read-source-not-flagged
  cases.
- Updated two pre-#28 tests that pinned the old
  `echo x > /etc/profile` => `unknown` / silent behavior to the new
  `unsafe` / `ask` result.
- Full suite: 407 tests green in a clean env (CI-equivalent).

## Out of scope

- `cp -t DIR SRC...` / `install` bare `SRC DST` dest detection: `cp`/`mv`
  use the common `SRC DST` (last-positional) form; `install` uses `-t`
  per the issue. The `-t`-with-sources form falls back to the generic
  `default: unsafe` reason (decision still correct).
- Symlinked paths (`/etc` -> `/private/etc` on macOS) match on the
  literal token, not the resolved path.
- Plugin manifest paths beyond `settings.json` / `settings.local.json`
  are not enumerated; the issue named those two concretely.

Closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,11 +90,20 @@ the AST. Visitor dispatch:
 - `if_statement`, `for_statement`, `while_statement`, `case_statement`,
   `do_group` — recurse into bodies. No manual keyword stripping required;
   the grammar already separates control-flow tokens from commands.
-- `redirected_statement` — check redirect targets against the
-  `safe_write_targets` glob list in `rules/shell.json` (defaults
+- `redirected_statement` — check redirect targets against two glob
+  lists in `rules/shell.json`. The `unsafe_write_targets` deny list
+  (dotfiles / config / startup paths like `~/.bashrc`,
+  `~/.ssh/authorized_keys`, `/etc/*`) is checked first: a match
+  classifies `unsafe` (ask with a specific reason). It is checked
+  *before* `safe_write_targets`, so a deny entry overrides a broader
+  safe glob — `~/.claude/settings.json` is unsafe even though
+  `~/.claude/*` is a safe-write target, because settings.json can
+  disable this hook. The `safe_write_targets` white list (defaults
   include `/dev/null`, `/tmp/*`, `/var/folders/*`, `~/.cache/*`,
-  `~/.claude/*`, etc.). Anything not on the list falls through to
-  `unknown` so Claude Code default-prompts. For
+  `~/.claude/*`, etc.) makes a write benign. Anything on neither list
+  falls through to `unknown` so Claude Code default-prompts. The same
+  deny list also routes the write-target arguments of
+  `tee` / `cp` / `mv` / `install` / `dd` / `find -fprint*`. For
   `python3 << ... <<EOF` heredocs, the body goes to the Python
   analyzer.
 - `command_substitution` (`$(...)`, `` `...` ``) and `process_substitution`
@@ -287,7 +296,9 @@ Example decisions (see `rules/shell.json` for the full rule set):
 | `echo foo \| xargs rm` / `echo foo \| xargs cat`             | ask / allow |
 | `time aws ec2 describe-instances`                            | allow    |
 | `cat file > /tmp/out`                                        | allow (`/tmp/*` is on the safe-write list) |
-| `cat file > /etc/profile`                                    | unknown (writes to a system path) |
+| `echo x > /etc/profile` / `echo x > ~/.bashrc`               | ask (on the `unsafe_write_targets` deny list) |
+| `echo x > ~/.claude/settings.json`                           | ask (deny list overrides the `~/.claude/*` safe glob) |
+| `echo x > ~/.claude/cache.json`                              | allow (`~/.claude/*` safe-write; not on the deny list) |
 | `aws ec2 describe-instances > /dev/null`                     | allow    |
 | `sqlite3 db.sqlite "SELECT * FROM t"`                        | allow    |
 | `sqlite3 db.sqlite "DROP TABLE t"`                           | ask      |
@@ -446,6 +457,13 @@ name rather than guessed.
     "/scratch/*"
   ],
 
+  "unsafe_write_targets": [
+    "~/.bashrc",
+    "~/.ssh/authorized_keys",
+    "/etc/*",
+    "~/.claude/settings.json"
+  ],
+
   "interpreters": {
     "python3": {
       "inline_flag": "-c",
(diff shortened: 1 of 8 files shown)


debedb:
[claude] Rebased onto `master` (now at f09d2c8) — the branch was conflicting after #47/#48 landed.

Conflicts were in `hooks/rule_classifier.py` and `tests/test_adversarial_regressions.py`. #48 renamed the *allowlist* terminology to *whitelist* while this PR generalized `_path_matches_safe_write_targets` into `_path_matches_target_list` (shared by the safe + deny checks); I took the generalization and adopted #48's "white list" wording in the PR-added prose and the `test_whitelist_still_overrides_unsafe_redirect` method name. The deny-list concept keeps the "deny list" naming (#48 introduced no opposing term).

CI is green on Python 3.10/3.11/3.12 and the PR is mergeable. Ready for review.

debedb (commented):
[codex] blocking: `_walk_redirected` stops after the first write redirect, so a safe first redirect masks later unsafe or unknown write targets. On this branch, `python3 hooks/grammar_classifier.py 'echo x > /tmp/safe > ~/.bashrc'` and `python3 hooks/grammar_classifier.py 'echo x > /tmp/safe 2> ~/.bashrc'` both return `{"decision": "safe", "reason": "builtin: echo"}`. The previous loop kept scanning all `file_redirect` nodes until it found a non-safe write; this needs to evaluate every write redirect with unsafe > unknown > safe precedence before falling through to the command.

Verification: targeted issue #28 tests pass (`python3 -m unittest tests.test_grammar_classifier.TestUnsafeWriteTargets tests.test_adversarial_regressions.TestIssue28UnsafeWriteTargets tests.test_rule_classifier.TestUnsafeWriteTargetArgs -v`), but the multi-redirect repro above is a regression and should be covered before merge.

debedb (commented):
[codex] LGTM. The follow-up commit fixes the multi-redirect masking regression by evaluating all write redirects with unsafe > unknown > safe precedence, and I verified the prior repros plus the targeted issue #28 suites.

debedb:
[claude] Addressed in ed0a67d. `_walk_redirected` now collects every write redirect target and applies unsafe > unknown > safe precedence instead of stopping at the first, so a safe leading redirect no longer masks a later protected or unknown one. Both repros (`echo x > /tmp/safe > ~/.bashrc` and the `2>` variant) now classify `unsafe`; all-safe statements still fall through to classify the command.

Regression coverage added in `TestUnsafeWriteTargets` (grammar level: stdout+stdout and stdout+stderr masking, unknown-precedence, all-safe control) and `TestIssue28UnsafeWriteTargets` (hook-level). The targeted #28 suite is green. Thanks for the catch and the re-verify.

Issue #46, opened 2026-06-07
Codex parity: friction-review loop for Codex CLI


debedb:
Split out of #44 (bullet: "This only applies to Claude Code. File and link issues for Codex to do the same").

The self-improvement loop in #44 is Claude Code specific in two ways: friction capture rides the PreToolUse/PostToolUse hook pair, and remediation writes `settings.json` `permissions.allow` globs. Codex CLI has neither surface in the same shape, so parity needs its own design.

What Codex offers instead (to be verified against current Codex CLI):

- `~/.codex/config.toml` — `approval_policy` (`untrusted` / `on-failure` / `on-request` / `never`) and `sandbox_mode` control when the user is prompted; there is no per-command-pattern allowlist comparable to `Bash(prefix:*)`.
- No PreToolUse-style hook API for vetoing/approving individual shell invocations, so YOLT cannot sit in the approval path the way it does in Claude Code.
- `notify` hook can observe events but not decide them.

Candidate directions:

1. **Capture**: wrap the shell entry point Codex uses (or parse Codex session logs) to record which commands triggered approval prompts — the equivalent of `~/.claude/yolt.log`.
2. **Review**: the `/yolt:review` generator is CLI-agnostic once the log exists; reuse it to produce the same markdown review doc.
3. **Remediation**: map accepted suggestions onto whatever Codex exposes — likely coarser (`approval_policy` per project trust) than Claude Code globs. Where Codex cannot express a rule, the YOLT classifier itself (running as a wrapper) is the fallback.

Design-only for now; depends on the v1 loop landing for Claude Code first (#44).

debedb:
Closing as part of a strategy change, not because the report is wrong.

Claude Code now ships auto mode — an LLM permission classifier on every tool call — as the **default** permission mode. YOLT is being narrowed to a deny-only guard over a short list of decisions we refuse to delegate to a probabilistic classifier, and will stop emitting `allow` entirely. Everything whose purpose is "make more commands classify as safe" is retired with it.

The premise here was that Codex has no auto-mode equivalent, so the friction-review loop retains full value there. On review that is building on sand: every vendor has the same capability and the same incentive, so a competitor's absence has a known expiry date. Retiring the rule-based half everywhere rather than everywhere-except-one.

Issue #45, opened 2026-06-07
Reviewer routing: write user overrides (~/.claude/yolt/*.json) for rules deeper than settings.json globs


debedb:
Split out of #44 (bullet: "custom settings that voitta-yolt can allow (that are obviously deeper than what `settings.json` allows)").

The `/yolt:review` loop (#44, v1 PR forthcoming) routes each friction candidate to one of three destinations:

| Candidate shape | Destination |
| --- | --- |
| Prefix-expressible, read-only regardless of flags (`aws logs get-log-events`) | `settings.json` `permissions.allow` — if settings.json can express it, it should |
| Flag-conditional / verb-class / compound-inner-only | `~/.claude/yolt/shell.json` or `~/.claude/yolt/rules.json` user override |
| Generic CLI gap useful to everyone | Upstream issue on this repo |

v1 of the reviewer surfaces paste-ready snippets for the first column and leaves the second column as manual edits. This issue tracks making the reviewer *write* the user-override files directly:

- Generate `~/.claude/yolt/shell.json` fragments (e.g. `commands.<cli>.safe_subcommands`, `nested_subcommand` specs, `unsafe_flags`) from observed friction, not just `Bash(...)` globs.
- Validate the merged result with `validate_shell_rules` before writing — a malformed override currently downgrades the whole hook to `rules-validation-error` / silent fallthrough, so the reviewer must never write an invalid file.
- Merge semantics: user override files merge per top-level key (overriding `safe_write_targets` replaces the list), so the writer has to read-modify-write rather than append.
- Decide inference depth: a repeated `unknown` on `mycli status` can be offered as `safe_subcommands: ["status"]`, but verb-class generalization (`describe-*` safe) should stay human-confirmed.

Why settings.json alone is not enough (the "deeper" part):

1. Claude Code allow patterns are exact/prefix only — no mid-pattern wildcard, so verb-class rules (`aws <any service> describe-*`) need one override line instead of hundreds of literal globs.
2. Flag-driven mutation (`gh api` vs `gh api -X POST`, `find` vs `find -delete`) cannot be distinguished by a prefix glob; `unsafe_flags` specs can.
3. A static settings.json allow bypasses the PreToolUse hook entirely, skipping redirect and command-substitution checks; a shell.json rule keeps the AST walk in the loop.

PR #47, opened 2026-06-08
Self-improvement review loop mining YOLT friction (#44)

### LINKED ISSUE #44: Self-improvement hook?
By analogy with [claudeception](https://github.com/blader/Claudeception) a potential hook that can surface all commands that required confirmation for user's review (as markdown document). Initially:

- Create local doc (w/o prompting)
- Ask user if they want to look at this and edit at all
- Then potentially submit as an issue to this repo.
- This also results in another issue (TO CREATE!) -- as to custom settings that voitta-yolt can allow (that are obviously deeper than what `settings.json` allows). 
- Conversely, if we can do things in `settings.json` (in `~/.claude`) we should
- This only applies to Claude Code. File and link issues for Codex to do the same.

This needs a bit of thinking through.

debedb:
## What

Self-improvement loop for YOLT (issue #44): mine the decision log for
recurring Bash friction and surface it as a reviewable doc with routed
remediations.

## Why

The dogfood log already records where YOLT got in the way. This turns
that record into action: which prefixes to allowlist, which need a local
rule, and which point at an upstream rules gap — without ever
recommending something that defeats YOLT.

## How

- **`hooks/yolt_review.py`** (stdlib only — works even when the
  tree-sitter bootstrap failed). Groups repeated commands to a
  conservative prefix and sorts them into three buckets:
  - `friction-unsafe` — YOLT returned `ask` repeatedly.
  - `friction-unknown` — fell through to Claude Code's default prompt.
  - `fastpath` — auto-allowed at high frequency (a `settings.json` glob
    would skip the hook startup).

  Writes `~/.claude/yolt/review.md` + `suggestions.json`;
  applied/dismissed status survives regeneration by suggestion id.

- **Glob-collision veto (safety-critical).** Before recommending any
  `Bash(prefix*)` allow glob, fnmatch it against every command YOLT did
  *not* classify safe. `gh api` is read-only but `gh api -X POST` is not,
  and both match `Bash(gh api*)` — promoting that glob would silently
  bypass YOLT for the POST. Any hit re-routes the suggestion to a
  `shell.json` rule (#45). Partially-overlapping namespaces stay
  suggestable (`gh pr view*` does not collide with `gh pr merge`).

- **Approval correlation.** A PostToolUse hook (`yolt_analyzer
  --ran-hook` -> `~/.claude/yolt-ran.log`) records commands that actually
  ran. A command YOLT said `ask` on only reaches PostToolUse if the user
  approved, so each friction suggestion carries an `approved` count.

- **Surfacing.** SessionStart prints a throttled (24h) nudge toward
  `/yolt:review`; SessionEnd regenerates with `--if-stale` (mtime-guarded
  no-op on quiet sessions). `/yolt:review` slash command triages and
  applies, honoring the routing above with user confirmation.

- **README** "Self-improvement loop" section; **41 stdlib-unittest
  tests** covering grouping, redaction, the collision gate, approval
  correlation, `build_groups`, and status-preserving regeneration.

## Testing

`python3 -m unittest discover tests` — all green under an isolated
`HOME`. (In a live shell, the 16 pre-existing end-to-end failures are the
real `~/.claude/settings.json` allowlist leaking into the hook tests, not
a regression here.)

## Privacy

Only the redacted `shape` field (argv head + flag names, values stripped)
may leave the machine in an upstream issue draft; raw `examples` stay
local.

Closes #44.
Refs #45 (write user overrides), #46 (Codex parity).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@
   - [Python rules — `~/.claude/yolt/rules.json`](#python-rules---claudeyoltrulesjson)
   - [Shell rules — `~/.claude/yolt/shell.json`](#shell-rules---claudeyoltshelljson)
 - [Debug / dogfood log](#debug--dogfood-log)
+- [Self-improvement loop](#self-improvement-loop)
 - [CLI usage](#cli-usage)
 - [Tests and demo](#tests-and-demo)
 - [Analysis boundaries](#analysis-boundaries)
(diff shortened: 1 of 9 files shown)


debedb (commented):
[codex] blocking: `was_approved` treats any later ran-record for the same command as approval for every earlier prompt. With prompts at t10 and t20 but only a ran-record at t21, both prompt records count as approved because the check only requires `ts >= record_ts - 5s`. That inflates `approved` and makes denied prompts look like friction the user accepted. Please consume/match ran records one-to-one or add a bounded correlation window so each PostToolUse event can only satisfy the corresponding prompt.

[codex] blocking: the glob-collision set includes the suggestion's own unsafe/unknown examples, so every `friction-unsafe` and `friction-unknown` suggestion collides with itself. For example, three `unknown` records for `kubectl get pods ...` generate `prefix="kubectl get pods"`, then `kubectl get pods ...` is already in `nonsafe_commands`, so `Bash(kubectl get pods*)` is always vetoed. That prevents the documented settings.json route for repeated safe-but-unknown/read-only commands and pushes all friction into shell rules/upstream issues.

Verification: `python3 -m unittest -v tests.test_yolt_review` passes, and `env HOME=/private/tmp/yolt-pr47-home python3 -m unittest discover tests` passes. Running the full suite under the live home reproduces the known 16 allowlist-leak failures from the real `~/.claude/settings.json`.

debedb:
Thanks @debedb — both blocking findings fixed in ef03ca9.

**1. `was_approved` inflation (one-to-many).** Replaced `was_approved`
with `correlate_approvals`, which matches prompts to ran-records
one-to-one per command: prompts are walked in time order, each consuming
the earliest still-unclaimed ran-record at/after `prompt_ts - skew`, so a
single PostToolUse event can credit at most one prompt. Your exact
scenario (prompts t10 + t20, lone ran t21) now yields `approved == 1`.
Covered by `test_one_ran_record_approves_only_one_prompt`. On the real
log, no suggestion now has `approved > fires`.

**2. Self-collision.** `annotate_glob_collisions` now takes the group's
own commands and excludes them from the scan. A friction suggestion's
own examples live in `nonsafe_commands` and the glob trivially matches
them, which was vetoing every `friction-unsafe` / `friction-unknown`
suggestion against itself. A genuinely *different* non-safe command swept
in by the glob still collides (e.g. fast-path `gh api` vs unsafe
`gh api -X POST` is preserved; `foo` vs `foo bar baz` still flagged).
Covered by `test_friction_unknown_readonly_does_not_self_collide`,
`test_own_group_commands_do_not_self_collide`, and
`test_distinct_command_still_collides_despite_own`. On the real log,
friction self-only collisions dropped from all-of-them to 0; the
fast-path veto still fires on 13/20 candidates.

Tests: +5 (now 46 in `test_yolt_review.py`). Full suite green under an
isolated `HOME`; the 16 live-shell failures remain the pre-existing
`~/.claude/settings.json` allowlist leak, unrelated to this change.

debedb (commented):
[codex] resolved: the self-collision fix addresses the routing bug. Excluding the group's own commands keeps read-only `friction-unknown` suggestions eligible for the documented settings.json route, while the tests still preserve collision detection for a different non-safe command swept in by the same glob.

[codex] blocking: `correlate_approvals` still has no upper bound on the ran-record match, so any later execution of the same command can back-credit an old denied prompt. The loop now consumes ran records one-to-one, but the match condition remains "first unclaimed ran at or after `prompt_ts - 5s`"; a denied prompt at t10 and a later run at t3600, after the command became statically allowed or otherwise bypassed prompting, still marks the t10 prompt approved. Please bound the match window around the prompt, or otherwise correlate to the nearest matching prompt rather than any future run for that command.

Verification: `python3 -m unittest -v tests.test_yolt_review` passes, `python3 -m unittest -v tests.test_yolt_review.TestApprovalCorrelation` passes, and `env HOME=/private/tmp/yolt-pr47-home python3 -m unittest discover tests` passes.

debedb:
Fixed in ae56ff5; thanks for catching the upper bound.

`correlate_approvals` now bounds the match to
`[prompt_ts - skew, prompt_ts + window]` with
`RAN_MATCH_WINDOW_SECONDS = 600`. The earliest still-unclaimed ran-record
inside that window approves the prompt and is consumed; a ran-record
beyond a prompt's window is left for a later prompt rather than
back-crediting an older one. So your scenario — a denied prompt at t0 and
a run 30 min later after the command became statically allowed — now
leaves the prompt unapproved.

Window choice: PostToolUse stamps completion time, so the bound has to
cover think-time plus command runtime while staying well below "a
separate later invocation." A genuinely-approved command slower than the
window is undercounted, which is the safe direction — better than
fabricating an approval. Documented on the constant.

Tests: +2 (`test_far_future_ran_does_not_backcredit_old_prompt`,
`test_ran_inside_window_approves`); 48 in `test_yolt_review.py`. Full
suite green under isolated `HOME`. Confirmed on the real log:
`approved > fires` count stays 0.

debedb (commented):
[codex] LGTM. The latest commit bounds approval correlation to a finite prompt window while preserving one-to-one ran-record consumption, and the prior self-collision routing issue remains resolved.

Verification: `python3 -m unittest -v tests.test_yolt_review` passes, `python3 -m unittest -v tests.test_yolt_review.TestApprovalCorrelation` passes, and `env HOME=/private/tmp/yolt-pr47-home python3 -m unittest discover tests` passes.

PR #49, opened 2026-06-09
Reviewer writes safe_subcommands overrides for personal CLIs (issue #45)

### LINKED ISSUE #45: Reviewer routing: write user overrides (~/.claude/yolt/*.json) for rules deeper than settings.json globs

debedb:
## Summary

Makes the `/yolt:review` reviewer **write** `~/.claude/yolt/shell.json`
overrides directly, instead of leaving every override as a manual
hand-edit. Split out of #44.

Adds a `--write-override <id>` subcommand to `hooks/yolt_review.py` that
writes the one fragment the reviewer can infer safely: a
`safe_subcommands` rule for a **personal CLI** (no bundled rule) that
repeatedly fell through to `unknown` on a subcommand.

```json
{"commands": {"mycli": {"default": "subcommand", "safe_subcommands": ["status"]}}}
```

`mycli status` then auto-allows while every other `mycli` subcommand
still prompts — the assertion is strictly additive and never flips an
existing `unsafe` verdict.

## What it does

- `build_groups` attaches an `override` decision to each
  `friction-unknown` suggestion (`writable` + `cli` + `fragment` +
  `label`, or a `reason`). Writable only for a personal CLI with a
  subcommand token; a `cli group sub` prefix nests under
  `nested_subcommand.<group>.safe_subcommands`.
- `--write-override` **read-modify-writes** the override file (unions the
  new subcommand, never clobbering existing entries), **validates** the
  merged result (bundled rules ∪ override) with `validate_shell_rules`,
  and only then writes via atomic temp+replace. A malformed merge is
  refused, so the writer can never downgrade the hook to
  `rules-validation-error`. Written suggestions are marked applied and
  the doc is regenerated.
- The override merge is extracted from `load_shell_rules` into
  `rule_classifier.merge_shell_overrides`, so the writer's validation
  runs the **exact** merge the hook performs (one source of truth).
- `review.md` / `README` document the route; the review doc now prints
  the concrete `--write-override` command and fragment.

## Decided scope (issue's "inference depth")

- **Personal-CLI only.** A `friction-unknown` on a *bundled* CLI is a
  rules gap worth an upstream issue, not a local shadow that freezes a
  stale copy of the bundled spec.
- **`safe_subcommands` only.** `friction-unsafe`, flag-conditional, and
  verb-class overrides stay human-authored — auto-flipping `unsafe`→`safe`
  is a security-regression vector the issue says must stay
  human-confirmed.

## Tests

`tests/test_yolt_review.py`: override eligibility (personal / nested /
bundled / bare / rules-unavailable), fragment shapes, merge
union/idempotence/no-clobber, and end-to-end `--write-override` including
validate-before-write refusal (file untouched), unknown / not-writable
id refusal, multi-id union, and a round-trip that loads the written
override through the real classifier. Full suite green under a clean
HOME (454 tests).

Closes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,9 +554,11 @@ Each suggestion routes to exactly one of three remediations:
   of flags. Fastest, but a static allow rule bypasses YOLT's hook
   entirely (including its redirect and command-substitution checks).
 - **Local override** — a `~/.claude/yolt/shell.json` rule for anything
-  flag-conditional or verb-class, keeping the AST walk in the loop
-  (generating these directly is issue
-  [#45](https://github.com/voitta-ai/voitta-yolt/issues/45)).
+  flag-conditional or verb-class, keeping the AST walk in the loop. For
+  the narrow, additive case of a personal CLI hitting `unknown` on a
+  subcommand, the reviewer writes the `safe_subcommands` override itself
+  (`--write-override <id>`); deeper rules stay hand-written. See
+  [Writing overrides](#writing-overrides-issue-45).
 - **Upstream issue** — a common CLI repeatedly hitting `unknown` is
   likely a rules gap worth reporting on voitta-ai/voitta-yolt.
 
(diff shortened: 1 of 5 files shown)


debedb (commented):
[codex] blocking: `--write-override` can still clobber an invalid existing per-command override before validation. In `cmd_write_override`, a non-dict `commands.<cli>` value is replaced with `{}` before `validate_shell_rules` runs, so an existing malformed user entry like `{"commands": {"mycli": "not-a-dict"}}` is silently overwritten, the command returns 0, and the suggestion is marked applied. That violates the PR/issue contract that malformed preexisting overrides are refused and untouched, and it defeats the “never clobber existing entries” safety claim. Please reject non-dict existing command specs before merging the fragment and add a regression test for that case.

Validation evidence: clean-HOME `python3 -m unittest discover -v tests` passes (454 tests), but the focused repro against PR head `4674c5b` returned 0 and rewrote the invalid `commands.mycli` value into the generated `safe_subcommands` dict.

debedb:
[claude] Addressed in c8fcd7e.

`cmd_write_override` now refuses the whole write (rc 1, file untouched, suggestion left pending) when any target CLI already has a non-dict `commands.<cli>` spec, instead of replacing it with `{}` before validation. Unrelated non-dict specs were already caught by the post-merge `validate_shell_rules` pass; this closes the target-CLI hole where the bad value was overwritten before validation could see it. Added a regression test with your exact `{"commands": {"mycli": "not-a-dict"}}` payload asserting rc 1, byte-identical file, and `status == pending`. Full suite green under clean HOME (455).

debedb (commented):
[codex] LGTM. Re-reviewed `c8fcd7e`; the prior `commands.<cli>` clobbering concern is resolved by refusing non-dict target command specs before mutation, and the added regression covers the exact failing payload.

Validation: `python3 -m unittest tests.test_yolt_review.TestWriteOverride.test_non_dict_existing_command_spec_is_refused_untouched -v`, `python3 -m unittest tests.test_yolt_review -v`, and clean-HOME `python3 -m unittest discover -v tests` all pass (455 tests).

PR #50, opened 2026-06-10
Detect SQL function side effects in keyword-only scanner (#26)

### LINKED ISSUE #26: SQL function side effects bypass keyword-only scanner
Split from #23.

## Problem

The SQL CLI scanner (README: SQL CLIs section) strips comments/literals, then keyword-matches against `INSERT/UPDATE/DELETE/DROP/...`. Leading `SELECT`/`WITH`/`EXPLAIN`/`SHOW` then classifies safe. This misses side-effecting *functions* called from inside a SELECT — the body looks like a read.

### Postgres
- `psql -c "SELECT pg_terminate_backend(12345)"` — kills a session.
- `psql -c "SELECT pg_cancel_backend(12345)"`.
- `psql -c "SELECT pg_promote()"` — failover replica.
- `psql -c "SELECT pg_reload_conf()"`, `SELECT pg_rotate_logfile()`.
- `psql -c "SELECT nextval('s')"` / `SELECT setval('s', 1)` — mutates sequence.
- `psql -c "SELECT lo_unlink(oid)"` / `SELECT lo_export(oid, '/path')` — delete large object / write file.
- `psql -c "SELECT pg_read_file('/etc/passwd')"` / `SELECT pg_ls_dir('/')` — FS exfil.
- `psql -c "SELECT dblink_exec('conn', 'DROP TABLE t')"` — remote DDL.
- `psql -c "SELECT pg_logical_emit_message(...)"` — writes WAL.
- `psql -c "SELECT pg_advisory_lock(1)"` — session lock.
- `psql -c "SELECT pg_sleep(1e9)"` — DoS.
- `psql -c "SELECT set_config('search_path', 'evil', false)"` — session state.
- `psql -c "DO \$\$ BEGIN PERFORM pg_terminate_backend(0); END \$\$"` — anonymous procedure body bypasses keyword check; `DO` is not in the destructive list today.

### MySQL / MariaDB
- `mysql -e "SELECT GET_LOCK('x', 9999)"`.
- `mysql -e "SELECT SLEEP(1e9)"` — DoS.
- `mysql -e "SELECT BENCHMARK(1e9, MD5('x'))"` — DoS.
- `mysql -e "SELECT LOAD_FILE('/etc/passwd')"` — exfil.
- `mysql -e "SELECT * FROM t INTO OUTFILE '/tmp/x'"` — verify `INTO` survives keyword strip (probably does, but should be in the test matrix).

### SQLite
- `sqlite3 db "SELECT load_extension('/path/to/evil.so')"` — native code execution.
- `sqlite3 db "SELECT randomblob(1<<40)"` — DoS.

### Edge cases
- `EXPLAIN ANALYZE INSERT INTO t ...` — Postgres actually executes the wrapped DML. Verify the inner verb is not masked by the leading `EXPLAIN`.
- `WITH d AS (DELETE FROM t RETURNING *) SELECT * FROM d` — Postgres allows DML in a CTE. Verify the scanner is not gated on the *first* keyword only.
- psql backslash meta in the same arg: `\copy`, `\i`, `\!`, `\o file`, `\g | cmd`, `\setenv` — verify coverage.

## Proposed approach

1. After comment/literal stripping, extract function-call identifiers via a lightweight scan (`IDENT(`).
2. Match against per-dialect deny lists (Postgres `pg_terminate_backend` etc., MySQL `LOAD_FILE`/`GET_LOCK`/`SLEEP`/`BENCHMARK`, SQLite `load_extension`/`randomblob`).
3. Match against per-dialect deny *prefixes* (Postgres `pg_*` system functions are a finite, mostly-mutating set; allow-list the read-only ones — `pg_database_size`, `pg_relation_size`, `pg_total_relation_size`, `pg_typeof`, `pg_get_*`, `pg_size_pretty`).
4. Add `DO` to the destructive keyword list.
5. Add tests for `EXPLAIN ANALYZE <dml>` and `WITH ... AS (DELETE ...) SELECT`.

Dialect is already known per CLI (`psql` → postgres, `mysql`/`mariadb` → mysql, `sqlite3` → sqlite, `duckdb` → duckdb), so per-dialect lists are cheap to dispatch on.

debedb:
## Summary

Closes #26.

The SQL CLI scanner stripped comments/literals, then keyword-matched against
`INSERT/UPDATE/DELETE/DROP/...`. A leading `SELECT/WITH/EXPLAIN/SHOW`
classified the statement safe — which missed side-effecting **functions**
called from inside a read-looking statement.

## What changed

`hooks/rule_classifier.py` — after the existing keyword pass in
`classify_sql_text`, scan `IDENT(` function calls against per-dialect deny
lists (dialect dispatched on the CLI name):

- **postgres** (`psql`): `pg_terminate_backend`, `pg_cancel_backend`,
  `pg_promote`, `pg_reload_conf`, `pg_read_file`, `pg_ls_dir`, `nextval`,
  `setval`, `lo_unlink`/`lo_export`/`lo_import`, `dblink_exec`,
  `pg_logical_emit_message`, `pg_advisory_lock`, `pg_sleep`, `set_config`, …
  plus a `pg_*` prefix deny with a read-only allowlist
  (`pg_database_size`, `pg_relation_size`, `pg_total_relation_size`,
  `pg_typeof`, `pg_get_*`, `pg_size_pretty`).
- **mysql / mariadb**: `load_file`, `get_lock`, `sleep`, `benchmark`, and the
  non-function `INTO OUTFILE` / `INTO DUMPFILE` file-write tokens.
- **sqlite3**: `load_extension`, `randomblob`.

Also added `DO` to `SQL_MUTATING_KEYWORDS` so anonymous procedure blocks
(`DO $$ ... $$`) no longer bypass the keyword check.

`EXPLAIN ANALYZE <dml>` and `WITH ... AS (DELETE ... RETURNING) SELECT` were
already caught by the scan-anywhere keyword pass; added regression tests.

Function names inside string literals stay safe (literals are stripped before
scanning).

## Tests

Added `TestSqlFunctionSideEffects` (30 cases) in
`tests/test_rule_classifier.py`. New SQL tests pass.

Note: the local suite shows 16 pre-existing failures under Python 3.14
(adversarial `find -exec` / `gh api` / `python3 -c` / import-alias and hook
tests). They are identical on clean `master` (verified by diffing the failure
set) — not introduced by this change; CI runs 3.10-3.12.

### DIFF
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -57,13 +57,21 @@ BASH_PATTERN_RE = re.compile(r"^Bash\((.*)\)$")
 # keyword. This treats `EXPLAIN DELETE FROM t` as unsafe (the DELETE
 # keyword is still present even though it would not execute); we accept
 # that false-positive to keep the rule simple.
+#
+# Keyword-only scanning misses side-effecting *functions* called from
+# inside an otherwise read-looking statement (`SELECT pg_terminate_backend(...)`,
+# `SELECT LOAD_FILE('/etc/passwd')`, `SELECT load_extension('evil.so')`).
+# After the keyword pass we therefore scan `IDENT(` function calls against
+# per-dialect deny lists, plus MySQL `INTO OUTFILE/DUMPFILE` file-write
+# tokens. Dialect is taken from the CLI name (psql->postgres, mysql/mariadb
+# ->mysql, sqlite3->sqlite, duckdb->duckdb). See issue #26.
 
 SQL_MUTATING_KEYWORDS = frozenset({
     "INSERT", "UPDATE", "DELETE", "DROP", "CREATE", "ALTER",
     "TRUNCATE", "REPLACE", "MERGE", "GRANT", "REVOKE",
     "VACUUM", "REINDEX", "ATTACH", "DETACH",
     "COMMIT", "ROLLBACK", "BEGIN", "SAVEPOINT", "RELEASE",
-    "CALL", "EXEC", "EXECUTE",
+    "CALL", "EXEC", "EXECUTE", "DO",
     "COPY", "LOAD", "IMPORT",
     "LOCK", "UNLOCK",
     "SET", "RESET",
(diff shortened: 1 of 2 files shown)


debedb (commented):
[codex] blocking:

- **blocking**: MySQL `OUTFILE`/`DUMPFILE` detection is too broad. The new token scan flags any occurrence of those identifiers, so safe reads like `SELECT outfile FROM t`, `SELECT dumpfile FROM t`, or `SELECT * FROM outfile` now classify unsafe. The issue asks to catch the file-write form `SELECT ... INTO OUTFILE/DUMPFILE`; this should require the `INTO` context rather than matching the token anywhere.
- **blocking**: Postgres dollar-quoted string literals are not stripped before the new function scan. `SELECT $$pg_sleep(9)$$` and `SELECT $tag$pg_sleep(9)$tag$` now classify unsafe even though the denied function name is inside a string literal. The PR body calls out that function names inside string literals stay safe, but the new regression test only covers single-quoted strings.

Validation run:

- `python3 -m unittest tests.test_rule_classifier.TestSqlFunctionSideEffects tests.test_rule_classifier.TestClassifySqlText tests.test_rule_classifier.TestParseSqlCliArgv` passed: 63 tests.
- `python3 -m unittest discover -v tests` reproduced the documented 16 unrelated failures outside the SQL classifier surface.

debedb:
[claude] Addressed in 5ecdfbe.

1. **OUTFILE/DUMPFILE scope** — replaced the bare-token set with
   `\bINTO\s+(?:OUTFILE|DUMPFILE)\b`, so `SELECT outfile FROM t`,
   `SELECT dumpfile FROM t`, and `SELECT * FROM outfile` now classify
   safe; `SELECT ... INTO OUTFILE/DUMPFILE '...'` stays unsafe.
2. **Dollar-quoted literals** — added `$$...$$` / `$tag$...$tag$`
   stripping to the literal strip regex, so `SELECT $$pg_sleep(9)$$`
   and the tagged form classify safe. `DO $$ ... $$` stays unsafe via
   the `DO` keyword (matched before the function scan).

Added regression tests for the tagged/untagged dollar-quote forms and
the bare outfile/dumpfile identifier forms. SQL classes pass (68);
full suite unchanged at the documented 16 unrelated 3.14 failures.

debedb (commented):
[codex] LGTM self-review.

Resolved in `5ecdfbe`: MySQL `OUTFILE`/`DUMPFILE` detection now requires the `INTO` context, and Postgres dollar-quoted literals are stripped before side-effecting function scanning. Targeted SQL tests pass at 68 cases; full suite remains at the documented 16 unrelated Python 3.14 failures.

PR #51, opened 2026-06-10
Recurse SQL scanner into cloud-CLI SQL flags (#29)

### LINKED ISSUE #29: Recurse SQL scanner into CLI flags carrying SQL strings
Split from #23.

## Problem

Several AWS / cloud CLIs accept a SQL (or SQL-shaped) string as a flag value. Today the classifier decides safe/unsafe by the verb only; the payload is opaque.

| Command | Flag carrying SQL | Current behavior |
| --- | --- | --- |
| `aws athena start-query-execution` | `--query-string` | unsafe via `start-*` |
| `aws rds-data execute-statement` | `--sql` | unsafe via `execute-*` |
| `aws timestream-query query` | `--query-string` | `query` not in either list → unknown |
| `aws logs start-query` | `--query-string` | safe via override (CWL Insights syntax cannot mutate; defensible but fragile) |
| `aws redshift-data execute-statement` | `--sql` | unsafe via `execute-*` |
| `gh api graphql -f query='mutation {...}'` | `-f query=` / `--field query=` | GET → safe; POST → unsafe via `-X POST` |

The verb-based safe/unsafe call is sound today. The risk is:
1. A user adds an `extra_safe_patterns` override for one of these (like the existing one for `aws logs start-query`) and the SQL is never scanned.
2. `aws timestream-query query` (verb `query`) doesn't match either pattern → unknown, prompts every time, even for trivially safe queries.

## Proposed approach

Define a small registry of (command, subcommand, flag) → `sql_scan` mappings. When the rule classifier sees a match, it pulls the flag value and runs the existing SQL scanner against it, using the same per-dialect ruleset as the corresponding direct CLI.

```json
{
  "aws": {
    "sql_payload_flags": {
      "athena start-query-execution": {"flag": "--query-string", "dialect": "presto"},
      "rds-data execute-statement":   {"flag": "--sql",          "dialect": "varies"},
      "timestream-query query":       {"flag": "--query-string", "dialect": "timestream"},
      "redshift-data execute-statement": {"flag": "--sql",       "dialect": "redshift"}
    }
  }
}
```

Decision precedence: the verb-based decision and the payload-based decision are AND'd toward unsafe — verb-unsafe stays unsafe; verb-safe + payload-unsafe → unsafe; verb-safe + payload-safe → safe.

## Why this matters even when verb already flags unsafe

It lets us *narrow* the allow surface without weakening it: a user who currently writes `Bash(aws athena start-query-execution --query-string "SELECT*")` to bypass the prompt could instead allowlist `aws athena start-query-execution*` and rely on payload scanning to keep destructive SQL flagged. Today they have no way to express "I trust read-only Athena queries but want destructive ones blocked."

## Related

Sibling issue on SQL function side effects feeds this: once the SQL scanner handles `pg_terminate_backend` / `load_extension` / etc., payload-scanned cloud-API queries inherit that coverage for free.

debedb:
## Summary

Recurse the existing SQL scanner into AWS CLI flags that carry a SQL
string, so the safe/unsafe call reflects the payload — not just the
verb. Adds a `sql_payload_flags` registry to the `aws` rule:

```json
"sql_payload_flags": {
  "athena start-query-execution":    {"flag": "--query-string", "dialect": "presto"},
  "rds-data execute-statement":      {"flag": "--sql",          "dialect": "varies"},
  "timestream-query query":          {"flag": "--query-string", "dialect": "timestream"},
  "redshift-data execute-statement": {"flag": "--sql",          "dialect": "redshift"}
}
```

When an aws operation matches, the flag value is pulled and run through
`classify_sql_text`.

## Decision precedence (the floor model)

The verb decision is a **floor** — payload scanning can only escalate or
refine, never weaken a mutating op on its own:

| verb decision | + payload | result |
| --- | --- | --- |
| `unsafe` (start-*, execute-*) | anything | `unsafe` |
| `safe` (via user override) | `safe` | `safe` |
| `safe` (via user override) | `unsafe` | `unsafe` |
| `safe` (via user override) | unclassified | `unknown` |
| `unknown` (timestream `query`) | payload | payload result |

This reconciles the two goals in the issue:

- **Out of the box**, only `aws timestream-query query` changes: verb
  `query` matched no pattern (→ `unknown`, prompted every time), and its
  payload now refines it to `safe`/`unsafe`.
- The `start-*`/`execute-*` ops stay `unsafe` by default. The narrowing
  use case ("I trust read-only Athena queries but want destructive ones
  blocked") is unlocked by the user adding an `extra_safe_patterns`
  override for the op — payload scanning then keeps destructive SQL
  flagged instead of the override blanket-allowing everything. That is
  risk #1 in the issue, now mitigated.

## Notes / scope decisions

- `classify_sql_text` gains an optional `dialect=` override (backward
  compatible) so the synthetic `cmd_name` used for payload scanning can
  still route the per-dialect function-side-effect scan from #26.
  `presto`/`timestream`/`redshift`/`varies` have no function deny-set
  yet, so only the dialect-agnostic keyword scan applies to them today;
  the field is recorded so a future deny-set lights up for free.
- `gh api graphql` is **not** added — its mutation/POST path is already
  `unsafe` via `-X POST`, as the issue notes; the proposed registry was
  aws-only.
- Validator learns `sql_payload_flags` (keys `flag`/`dialect`, `flag`
  required, underscore-prefixed meta keys skipped) so rule drift still
  fails at load time.

## Tests

+23 tests, all green; full suite has no new failures vs. the base
(the 16 pre-existing failures are the known Python-3.14-only phantom
set, green on CI 3.10-3.12).

- `extract_flag_value`: space / equals / value-with-equals / absent /
  flag-last forms.
- combine matrix: timestream SELECT→safe, DELETE→unsafe, missing
  flag→unknown; write-verb floor for athena/rds-data/redshift-data with
  no override; override + SELECT→safe, DROP→unsafe, unclassified→unknown;
  unregistered op untouched.
- dialect-param routing (sqlite deny-set via explicit dialect vs. no-op
  when None).
- validator drift (unknown key, missing flag, `_note` allowed).
- e2e through the grammar classifier (timestream safe/unsafe, athena
  floor).

Closes #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,37 @@ are classified separately by name — `.tables` / `.schema` / `.headers`
 are safe; `.import` / `.load` / `.read` / `.shell` / `.backup` are
 unsafe.
 
+### SQL carried in cloud-CLI flags
+
+Several AWS CLIs take a SQL string as a flag value (`aws athena
+start-query-execution --query-string`, `aws rds-data execute-statement
+--sql`, `aws timestream-query query --query-string`, `aws redshift-data
+execute-statement --sql`). The `aws` rule names these in a
+`sql_payload_flags` registry keyed by `"<service> <operation>"`; each
+entry gives the SQL-carrying `flag` and a `dialect`. When the operation
+matches, the flag value is pulled and run through the same SQL scanner
+described above.
+
+The verb decision is a **floor**, so payload scanning never weakens a
+mutating operation on its own:
+
+- A write verb (`start-*`, `execute-*`) stays `unsafe` regardless of
+  payload. Reading the SQL only helps once the user has explicitly
+  marked the operation safe — e.g. an `extra_safe_patterns` override
+  for `start-query-execution`. After that, the payload governs:
+  read-only `SELECT` → `safe`, destructive SQL → `unsafe`, unclassified
+  SQL → `unknown`. This is the point of the registry: an override that
+  used to blanket-allow every query now keeps destructive ones flagged.
+- `aws timestream-query query` matches no verb pattern (so it was
+  `unknown` and prompted every time); its payload now refines it to
+  `safe` / `unsafe` out of the box.
+
+`dialect` feeds the per-dialect function-side-effect scanner (see issue
+#26). `presto` / `timestream` / `redshift` / `varies` have no function
+deny-set yet, so only the dialect-agnostic keyword scan applies to them
+today; the field is recorded so adding a deny-set later lights up
+function detection for those services with no further wiring.
+
 ## Python rules (interpreter delegate)
 
 When the grammar walker hands a Python source body to the analyzer
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -1190,9 +1190,15 @@ class RuleClassifier:
         flag = entry.get("flag")
         sql = extract_flag_value(cmd_args, flag)
         if sql is None:
-            retval = (verb_decision,
-                      "{}; {}: no {} value to scan".format(
-                          verb_reason, label, flag))
+            # The operation is registered as SQL-bearing but the payload is
+            # not on the expected flag — absent, interactive, or supplied via
+            # an alternate input form such as `--cli-input-json`. The scanner
+            # cannot inspect it, so a verb-safe override must NOT blanket-allow
+            # the call; downgrade to unknown so Claude Code prompts. (A verb
+            # already classified unsafe returned above.)
+            retval = (DECISION_UNKNOWN,
+                      "{}: registered SQL op but no {} payload to scan; "
+                      "not certifying safe".format(label, flag))
             return retval
 
         dialect = entry.get("dialect")
--- a/hooks/rule_classifier.py
+++ b/hooks/rule_classifier.py
@@ -192,10 +192,16 @@ def _classify_sql_functions(cmd_name, dialect, stripped):
     return None
 
 
-def classify_sql_text(cmd_name, sql):
+def classify_sql_text(cmd_name, sql, dialect=None):
     """Return (decision, reason) for a single SQL string passed to a
     SQL CLI. Handles sqlite3 dot-commands too — they're not SQL but they
-    arrive through the same channel (positional / -cmd / -e)."""
+    arrive through the same channel (positional / -cmd / -e).
+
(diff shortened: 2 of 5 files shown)


debedb (commented):
[codex] blocking: `aws` SQL-payload entries can still become blanket-safe when the SQL flag is absent. With an `extra_safe_patterns` override for `athena start-query-execution`, `_scan_aws_sql_payload` returns the prior safe verb decision when `extract_flag_value` finds no `--query-string` value, so `aws athena start-query-execution` classifies `safe` with the reason `service override safe; ... no --query-string value to scan`. That breaks issue #29's goal of replacing a blanket allow with payload scanning: the scanner cannot inspect the payload, but the registered operation is still allowed. This also leaves alternate AWS input forms such as `--cli-input-json` outside the scanned flag path while inheriting the safe override. Please make the missing-payload case `unknown` for registered SQL-payload operations when the verb is not already unsafe, and add the corresponding override regression test.

Verification: `python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` passes on the PR worktree (`323` tests). A focused probe against PR head `634f4d7` reproduced the bypass above.

debedb:
[claude] Addressed in fb3fb06.

`_scan_aws_sql_payload` now returns `unknown` (not the prior safe verb decision) whenever a registered SQL-payload op has no value on its scanned flag and the verb is not already unsafe. So `aws athena start-query-execution` with a safe override but no `--query-string` — including the `--cli-input-json` form you called out — no longer certifies safe; the scanner can't inspect the payload, so Claude Code prompts. Added two override regression tests (missing flag, `--cli-input-json`); `python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` passes at 325 tests.

debedb (commented):
[codex] LGTM. The follow-up commit `fb3fb06` resolves the prior blocking finding: registered AWS SQL-payload operations now return `unknown` when the configured SQL flag is absent, so a safe override no longer blanket-allows an unscannable payload path such as `--cli-input-json`.

Verification: `python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` passes on the PR worktree (`325` tests). `python3 -m unittest discover -v tests` still shows the known 16 pre-existing environment/user-allow-pattern failures noted in the PR, with the new SQL-payload tests passing.

PR #57, opened 2026-06-12
feat: add eslint rule to shell classifier

debedb:
Adds an `eslint` rule to `rules/shell.json`: default safe (lint/report); `--fix`/`--fix-type` unsafe; `-o`/`--output-file` write-target; `--fix-dry-run` safe. Surfaced by double-door/doubledoor2026#75 (bare `eslint` gated as unknown).

### DIFF
--- a/rules/shell.json
+++ b/rules/shell.json
@@ -383,6 +383,13 @@
       "_note": "Removed 'tag' from flat safe/unsafe lists. It now lives under nested_subcommand.tag with positional-aware classification: bare 'git tag' lists safely; 'git tag v1' is unknown (not silently safe); mutating flags are explicit."
     },
 
+    "eslint": {
+      "default": "safe",
+      "unsafe_flags_without_value": ["--fix"],
+      "unsafe_flag_any_value": ["--fix-type"],
+      "write_flag_value_targets": ["-o", "--output-file"],
+      "_note": "eslint is read-only (lint/report) unless it rewrites files: --fix / --fix-type, or -o/--output-file writing outside safe_write_targets. --fix-dry-run stays safe."
+    },
     "docker": {
       "default": "subcommand",
       "safe_subcommands": [
--- a/tests/test_rule_classifier.py
+++ b/tests/test_rule_classifier.py
@@ -1137,5 +1137,54 @@ class TestLoadShellRulesValidation(unittest.TestCase):
         self.assertIn("commands", rules)
 
 
+class TestEslintRule(unittest.TestCase):
+    """End-to-end regression coverage for the eslint rule in rules/shell.json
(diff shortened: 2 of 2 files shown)


debedb (commented):
[codex] blocking: please add regression tests for the new `eslint` rule. The PR changes safety classification for a command that previously prompted as unknown, but the branch only updates `rules/shell.json`; CI would not catch a later drift where `eslint` becomes unknown again or where `--fix`, `--fix-dry-run`, or `--output-file` are misclassified. At minimum, cover bare/targeted `eslint` as safe, `--fix` as unsafe, `--fix-dry-run` as safe, and `-o` / `--output-file` with safe vs non-safe write targets.

Verification: I read the linked scope from double-door/doubledoor2026#75 and PR #57 context; there are no prior reviews, PR comments, or inline threads. On PR head `625eb71`, `python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` passes (`325` tests), and a local probe confirmed the intended current behavior for bare `eslint`, `--fix`, `--fix=true`, `--fix-dry-run`, `--fix-type`, and output-file targets. The blocker is that those eslint-specific expectations are not committed as tests.

debedb:
Addressed the blocking ask in 5753d52: added `TestEslintRule` to `tests/test_rule_classifier.py` — end-to-end regression coverage that loads the real `rules/shell.json` and classifies `eslint` argv, so a later drift (eslint regressing to unknown, or `--fix` / `--fix-dry-run` / `--output-file` misclassified) fails CI.

Cases, mapping to your requested list:
- bare `eslint`, `eslint src/app.js` -> safe
- `--fix`, `--fix=true` -> unsafe
- `--fix-dry-run` -> safe (asserts it is not swept up by the `--fix` prefix)
- `--fix-type suggestion` -> unsafe (any value)
- `-o /tmp/report.json`, `--output-file /tmp/report.json` -> safe (allow-listed target)
- `-o report.json`, `--output-file=report.json` -> unsafe (non-allow-listed path)

`python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` -> 335 tests OK (was 325; +10).

debedb (commented):
[codex] LGTM. Commit `5753d52` resolves the prior blocking finding by adding committed regression coverage for the new `eslint` rule, using the real `rules/shell.json` and covering bare/targeted lint, `--fix`, `--fix=true`, `--fix-dry-run`, `--fix-type`, and safe vs non-safe `-o` / `--output-file` targets.

Verification: re-read PR #57, linked issue double-door/doubledoor2026#75, prior reviews, PR comments, inline threads, and the new commit. On extracted PR head `5753d52`, `python3 -m unittest tests.test_rule_classifier tests.test_grammar_classifier` passes (`335` tests). A direct classifier probe also matched the expected eslint decisions.

PR #81, opened 2026-08-08
feat(rules): classify python test-runner invocations as read-only

debedb:
## Problem

These prompt on every run:

```
python3 -m unittest test_scan_transcripts -v          -> unknown
python3 -m unittest tests.test_mod.TestCase.test_x    -> unknown
python3 -m pytest -q tests/                           -> unknown
pytest -q                                             -> unknown
python -m pytest                                      -> unknown
```

`unittest`'s nested-module spec can only enumerate verbs, but the token after `-m unittest` is a test *target* (module / class / method path). `pytest` had no rule in any form. The `python` interpreter spec had no module policy at all, so venv `python -m ...` never matched.

## Changes

- **`_classify_nested_module` now implements `safe_subcommand_patterns` / `unsafe_subcommand_patterns`.** The schema validator already accepted both keys while the classifier silently ignored them — the exact drift `validate_shell_rules` exists to catch. Unsafe patterns are consulted first so a broad safe pattern cannot swallow a narrower mutating one.
- **New `inherits` key on interpreter specs**, with `python` inheriting `python3`. One level only: a base that itself inherits is left unexpanded (no cycles), and an unknown base is a validation error. Avoids a second copy of the module policy that drifts from the first.
- **`rules/shell.json`** — `safe_subcommand_patterns: ["*"]` on `unittest`, a new `pytest` nested module, a top-level `pytest` command.
- **README** — new "Test runners" section (+ TOC, boundaries table row).

After:

```
python3 -m unittest test_scan_transcripts -v   -> safe (matches safe pattern)
python3 -m pytest -q tests/                    -> safe
pytest -q                                      -> safe
python -m pytest                               -> safe (via inherits)
python -m pip install foo                      -> unsafe   (unchanged)
python -m http.server                          -> unsafe   (unchanged)
python3 -m nonesuch                            -> unknown  (unchanged)
```

## The policy call

A test run executes project code; a test is free to write files or call an API. `safe` here means *YOLT will not prompt*, not *provably no side effects*. That trade was already made for bare `python3 -m unittest` and `python3 -m unittest discover`; this extends it to the forms people actually type. The README states it plainly and shows the `~/.claude/yolt/shell.json` override that narrows it back.

Runners for other ecosystems (`go test`, `cargo test`, `npm test`) are deliberately **not** covered — `npm test` in particular runs whatever `package.json` says, which is often a build.

## Version

`0.2.0` → `0.3.0`. Minor per CLAUDE.md: user-visible classification change plus a new rules-file key.

## Testing

`python3 -m unittest discover tests`: 582 tests, same 3 environment-dependent failures as `master` in this environment (master shows 4). No new failures. Per repo convention no new tests were added in this PR.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "yolt",
   "description": "Static safety analyzer for Bash invocations - auto-allows read-only commands and prompts for mutating ones, plus a Python AST analyzer for python3 / python3 -c / heredoc invocations.",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "author": {
     "name": "voitta-ai"
   },
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@
 - [Dependencies](#dependencies)
 - [What the grammar classifier handles](#what-the-grammar-classifier-handles)
 - [SQL CLIs](#sql-clis)
+- [Test runners](#test-runners)
+  - [Interpreter aliases (`inherits`)](#interpreter-aliases-inherits)
 - [Python rules (interpreter delegate)](#python-rules-interpreter-delegate)
 - [Custom rules](#custom-rules)
   - [Python rules — `~/.claude/yolt/rules.json`](#python-rules---claudeyoltrulesjson)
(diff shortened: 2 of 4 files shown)


debedb:
Closing as part of a strategy change, not because the work is wrong — the `inherits` mechanism and the `_classify_nested_module` pattern fix are both clean, and the drift they caught (schema accepting keys the classifier ignored) was a real bug.

Claude Code now ships auto mode — an LLM permission classifier on every tool call — as the **default** permission mode. YOLT is being narrowed to a deny-only guard over a short list of decisions we refuse to delegate to a classifier, and will stop emitting `allow`. "Teach it that pytest is read-only" is the exact class of change being retired: auto mode reaches that conclusion with no rule entry and no maintenance.

If the deny-only rewrite ever needs nested-module handling for a *blocking* decision, this branch is the reference.

PR #82, opened 2026-08-09
feat(policies): allow self-authored commits, pushes, PRs, and comments

debedb:
## Why

The rules answer *"does this command mutate anything?"*. That question has no good answer for `git commit`: committing to a shared branch other people are building on is a very different act from committing to a feature branch you opened five minutes ago in your own worktree — and the argv is identical. So `git commit` is `unsafe`, and the same prompt gets approved all day.

This adds a **policy layer** that answers a second question: *"is this write confined to work that is already mine?"* That needs repository state, so it shells out to read-only `git`.

## The design-principle carve-out

"Purely syntactic; no subprocess fork" is not repealed, it is scoped:

- Probes run **only** after the static rules already returned `unsafe` **and** the argv head matched a configured policy entry. The common path never forks.
- A policy can **only** upgrade `unsafe` → `safe`. It never makes a safe command unsafe, and it never sees a command the rules cleared.
- **Every probe failure means "no"** — no git, not a repo, 3s timeout, unparseable output, a `-C` path containing a shell expansion. The original `ask` stands.

## What is allowed by default

`self_authored_git`, enabled out of the box:

| Command | Requires |
| ------- | -------- |
| `git commit` | `linked_worktree` + `non_default_branch` + `solo_author` |
| `git push` (no force / delete / foreign refspec) | same |
| `gh pr create` | same |
| `gh issue comment`, `gh pr comment` | nothing — additive and reversible, no repository state |

Predicates:

- **`linked_worktree`** — `--git-dir` != `--git-common-dir`, i.e. not the primary checkout. Under the repo-stays-on-default-branch + `<repo>.worktrees/<branch>` convention, this alone says "this is branch work".
- **`non_default_branch`** — default from `origin/HEAD`, falling back to whichever of `init.defaultBranch` / `main` / `master` actually resolves *in this repo*. `init.defaultBranch` is a global preference for repos yet to be created, so it is treated as a candidate, never as an answer — otherwise setting it to `main` makes every existing `master` repo look like it has no default branch. Detached HEAD never qualifies.
- **`solo_author`** — every commit in `<default>..HEAD` was authored **and** committed by `git config user.email`. A branch with no commits of its own is vacuously solo — the state you are in the moment before the first `git commit`, which is exactly the case this exists to allow.

## Judging the right directory

`cd /path/to/wt && git commit` is the shape this exists for, so the walker tracks literal `cd` targets in source order, and `git -C <path>` retargets the probe. A `cd` that cannot be resolved statically (`cd "$DIR"`, `cd -`) sets the tracked directory to `None`, which disables the policy for the rest of the invocation rather than letting it judge the wrong tree.

## Verified matrix

In a real linked worktree on a solo branch:

```
git commit -m "x"                              safe   linked worktree on feature/..., authored only by you
git --no-pager commit -m "x"                   safe
git push                                       safe
git push -u origin feature/git-worktree-allow  safe
gh pr create --title t --body b                safe
gh issue comment 5 --body hi                   safe   allowed by policy
```

Still prompts:

```
git push origin master     unsafe   refspec the predicates never examined
git push --force           unsafe   deny_flags
cd $SOMEWHERE && git commit unsafe  unresolvable cd
cd /tmp && git commit      unsafe   not a repo
```

Against a synthetic repo (primary + linked worktree, then a commit forged with another `GIT_AUTHOR_EMAIL`):

```
worktree, solo branch                      safe
same branch after a commit by someone else unsafe
primary checkout on main                   unsafe
detached HEAD in the worktree              unsafe
YOLT_NO_POLICIES=1                         unsafe
```

## Configuration

Data like every other rule — `policies` in `rules/shell.json`, overridable from `~/.claude/yolt/shell.json`, `YOLT_NO_POLICIES=1` as a wholesale kill switch. Entry fields: `argv` (prefix, matched skipping flags, `argv[0]` by basename), `require`, `deny_flags`, `deny_default_branch_refspec`. Unknown keys and unknown predicate names fail schema validation at hook load — a typo here would be a false-*allow*, not a false-prompt.

`git add` is deliberately **not** in the defaults (staging is not what the policy was asked to cover), so `git add -A && git commit && git push` still prompts on the `git add`. The README shows the two-line override that adds it.

## Cost

0ms on every command that does not match. ~40ms on a match, memoized per directory so `git commit ... && git push` probes once.

## Version

`0.2.0` → `0.3.0`. Minor per CLAUDE.md: user-visible decision change, new config key, new env var. (Note: #81 also targets `0.3.0`; whichever merges second needs the trivial bump conflict resolved.)

## Testing

`python3 -m unittest discover tests`: 582 tests, same 3 environment-dependent failures as `master` here. No new failures. Per repo convention no new tests were added in this PR.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "yolt",
   "description": "Static safety analyzer for Bash invocations - auto-allows read-only commands and prompts for mutating ones, plus a Python AST analyzer for python3 / python3 -c / heredoc invocations.",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "author": {
     "name": "voitta-ai"
   },
--- a/README.md
+++ b/README.md
@@ -23,6 +23,10 @@
 - [What the grammar classifier handles](#what-the-grammar-classifier-handles)
 - [SQL CLIs](#sql-clis)
 - [Python rules (interpreter delegate)](#python-rules-interpreter-delegate)
+- [Policies](#policies)
+  - [What is allowed by default](#what-is-allowed-by-default)
+  - [Cost](#cost)
+  - [Turning it off or changing it](#turning-it-off-or-changing-it)
 - [Custom rules](#custom-rules)
   - [Python rules — `~/.claude/yolt/rules.json`](#python-rules---claudeyoltrulesjson)
   - [Shell rules — `~/.claude/yolt/shell.json`](#shell-rules---claudeyoltshelljson)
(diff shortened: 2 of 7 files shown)


debedb:
Closing the **direction**, not the work. The predicates in
`hooks/git_policy.py` are being kept and carried into #97.

## What changed underneath this PR

Claude Code now ships auto mode — an LLM permission classifier on every tool
call — as the **default** permission mode. YOLT is being narrowed to a
deny-only guard over a short list of decisions we refuse to delegate to a
probabilistic classifier, and will stop emitting `allow` entirely.

That makes this PR wrong-direction by construction: its entire function is
to upgrade `unsafe` -> `safe`, and it pays a subprocess fork per gated
command to grant approvals the platform now grants for free. It is also
approval-granting machinery that the platform's own tooling cannot see —
`/auto-mode-setup` hunts and deletes broad `permissions.allow` entries as
"classifier-bypassing", but a hook returning `allow` is the same bypass and
is invisible to it.

None of that was knowable when this was written.

## What this PR got right, on the record

**The diagnosis is the best sentence in the repo and it survives the
strategy change:**

> *"does this command mutate anything?"* has no good answer for `git commit`.

Committing to a shared branch and committing to a worktree branch you opened
five minutes ago are different acts with identical argv. No purely syntactic
classifier separates them, ever. Answering it needs repository state. That
was a correct read of a real ceiling, not a workaround.

**The engineering is disciplined in ways that deserve saying out loud:**

- Every probe failure means "no" — no git, not a repo, 3s timeout,
  unparseable output, a `-C` path containing a shell expansion.
- Can only loosen, never tighten; never sees a command the rules cleared.
- 0ms on non-match, ~40ms memoized per directory on match.
- Unknown keys and unknown predicate names fail schema validation **at hook
  load**, explicitly reasoned as "a typo here would be a false-*allow*, not
  a false-prompt."
- `init.defaultBranch` treated as a candidate and never an answer, because
  it is a global preference for repos that do not exist yet — otherwise
  every existing `master` repo looks defaultless. Subtle and correct.
- Detached HEAD never qualifies. `git add` deliberately excluded. A branch
  with no commits of its own is vacuously solo, which is exactly the state
  the moment before the first commit.
- The design-principle carve-out was argued rather than assumed.

The honest critique is not that any of this is wrong. It is that the
sophistication was itself the signal: when it takes three predicates, a `cd`
tracker and a fail-closed matrix to decide one command, the question is not
syntactic — which this PR says in its own opening line. It correctly found
the classifier's ceiling and then extended the classifier, rather than
concluding the classifier was the wrong instrument.

## What carries forward

#97 keeps all three predicates — `linked_worktree`, `non_default_branch`,
`solo_author` — and inverts them. Same probes, opposite conclusion: block a
push to the default branch, block a force-push against a branch carrying
commits that are not yours, optionally block a commit from the primary
checkout where the convention is worktree-per-branch.

Those are deterministic, reproducible, auditable facts. "Did someone else
author commits on this branch" is exactly the kind of question we should not
want answered probabilistically — so these become the **first concrete
entries in the non-delegable list**, which until now had no content at all.

One thing #97 flags that cannot be ported verbatim: inverting the direction
inverts the meaning of fail-closed. Here a probe failure meant *do not
grant*, which degraded toward friction and was safe. Deny-side, the naive
reading becomes *deny*, and a git outage or an unresolvable `cd` would brick
the session. The deny-side version must fall through to `unknown` on probe
failure.

The branch `feature/git-worktree-allow` stays as the reference
implementation. Not deleting it.

Issue #84, opened 2026-08-14
YOLT's own logs persist credentials verbatim (yolt.log / yolt-ran.log)


debedb:
## Summary

Both YOLT logs store the **full command string verbatim**, default-on, with no
redaction and no rotation. Any credential that appears in a command — which is
routine, since headers, tokens and connection strings are passed on command
lines — is persisted indefinitely in plaintext under `~/.claude/`.

This is not hypothetical. During a credential-leak cleanup on one machine, both
YOLT logs were found holding live secrets and had to be scrubbed:

| file | records | what was in it |
|---|---|---|
| `~/.claude/yolt.log` | 5141 | a **production API key** (fetched legitimately from a secret manager, then used in a `curl -H` on the next line) |
| `~/.claude/yolt-ran.log` | 3146 | the same key, plus chat-platform bot tokens |

The schemas make it unavoidable:

```
yolt.log      {agent_id, command, decision, permission_mode, reason, ts}
yolt-ran.log  {command, ts}
```

`command` is the raw string.

## Why this matters more than a normal log

Three properties compound:

1. **Default-on.** `YOLT_LOG_FILE` / `YOLT_RAN_LOG_FILE` unset means logging happens; opting out requires knowing to set them to empty.
2. **No rotation** — append-only and unbounded, so a token from months ago is still there long after it was rotated and believed gone.
3. **It is a security tool.** Users install YOLT to be *safer*. A safety hook that quietly accumulates a credential store is the least expected place for one, and therefore the least likely to be swept.

A leaked credential is normally found by grepping the obvious places. `~/.claude/yolt.log` is not on anyone's list.

## Suggested fix

Redact credential-shaped substrings in `command` before writing, in both log
paths. A command line is a *good* target for this — unlike prose, it is short
and structured, so prefix matching is accurate rather than best-effort. Minimum
useful pattern set:

```
gh[posru]_   github_pat_   glpat-   xox[baprse]-   xapp-
AKIA   ASIA   sk-   -----BEGIN * PRIVATE KEY-----
scheme://user:secret@host
```

plus assignment-shaped values (`--token X`, `-H "X-Api-Key: X"`, `TOKEN=X`),
where replacing only the value keeps the log useful for its actual purpose —
`yolt_review.py` mines these for recurring command *shapes*, and a redacted
value does not harm that.

Two details worth building in:

- **Redact at write time, not read time.** The risk is the file on disk.
[...]

PR #87, opened 2026-08-14
feat(hook): warn when a command line carries a credential (#85)

### LINKED ISSUE #85: Detect credentials in command strings at PreToolUse (advisory)
## Summary

YOLT already inspects every Bash command before it runs and classifies it for
safety. That makes it the natural place to catch a different class of hazard it
does not currently look for: **a credential embedded in the command string.**

Proposal: add a classification for commands carrying credential-shaped
substrings, advisory at first.

## Why this is the right hook point, and why it is tractable

A secret that appears once in an agent session does not land in one place. It
lands in the transcript, any session-memory store (usually embedded and
searchable, so it can resurface in a *later* session's context), spilled
tool-output files, the agent's logs, and the permission allowlist if the
approved command string contained it. Copies with different lifetimes, none in
git, most never swept.

`PreToolUse` is the **only** point where that is preventable rather than
cleanable. Once the command has run, every one of those copies already exists.

And crucially, this is the *easy* half of secret detection. Redacting secrets
out of prose is close to unwinnable — a single 64-char key was recently found
written 14 times in one document in 8 different phrasings, defeating the obvious
keyword-adjacency pattern in 6 of them (backticks, intervening words, markdown
emphasis, no delimiter at all). A **command line is not prose**: it is short,
structured, and secrets in it appear in known shapes. Prefix matching, which is
unreliable across a transcript, is accurate against `argv`.

YOLT already has the parsing machinery (`rule_classifier.py`,
`grammar_classifier.py`), already sees the command, and already has a decision
vocabulary to hang this on.

## What it would catch

The concrete case that motivated this: a key fetched *correctly* from a secret
manager, then used on the next line as

```
curl -H "X-Api-Key: <value>" https://service/endpoint
```

Nothing about that command is unsafe by YOLT's current lights, and it is not
wrong. But the string itself then reached four separate files on disk. A warning
at that moment — with the fix suggested — would have prevented all of it.

## Suggested shape

- **Advisory before blocking.** A control that false-positives on legitimate
  work gets disabled and then protects nothing. Warn, count, tighten later.
- **Suggest the remediation, do not just refuse.** The fix is mechanical: move
  the value into the environment so it never appears in `argv`.
  ```
  KEY="$(fetch)" sh -c 'curl -H "X-Api-Key: $KEY" https://service/endpoint'
  ```
  A hook that prints the corrected form is far likelier to be kept than one that
  only says no.
- **Structured prefixes first** — `gh[posru]_`, `github_pat_`, `glpat-`,
  `xox[baprse]-`, `xapp-`, `AKIA`/`ASIA`, `sk-`, PEM blocks,
  `scheme://user:secret@host`. Assignment shapes (`--token X`,
  `-H "Authorization: X"`) second, with a value-shape guard so that
  `--token $MY_TOKEN` and `--token some-resource-name` do not trip it.
- **Never log the matched value**, including in the warning text. Report shape
  and position.
- Worth noting `argv` is world-readable to anything that can run `ps` while the
  process lives, so this is not only about logs.

## Related

Companion to the log-redaction issue: that one stops YOLT persisting secrets it
already sees; this one stops the secret reaching the command line in the first
place. They are independent — the redaction fix is worth doing even if this is
never built.

debedb:
Closes #85.

> **Stacked on #86** (issue #84). Base is `feature/84-redact-secrets-in-logs`,
> so the diff shown here is #85's own change. Retarget to `master` once #86
> merges. The dependency is real: this reuses `find_secrets()` from #86.

## Why here

YOLT already parses every Bash command for safety, so `PreToolUse` is
the natural place to catch a hazard it was blind to — a credential in
the command string itself.

The motivating case is not a mistake:

```
curl -H "X-Api-Key: <value>" https://service/endpoint
```

The key was fetched correctly from a secret manager. Nothing about the
command is unsafe by YOLT's current lights. But that string then lands
in the transcript, in session memory (embedded and searchable, so it can
resurface in a *later* session), in spilled tool output, in YOLT's own
logs, and in `permissions.allow` if the approved command contained it.
Copies with different lifetimes, none in git, most never swept — plus
`argv` is readable by anything that can run `ps` while the process lives.

`PreToolUse` is the only point where that is preventable rather than
cleanable. Once the command has run, every one of those copies exists.

## Shape

Reuses `find_secrets()` from #86, so the match set and the
false-positive guards are shared — one place to tune, and structured
prefixes come first with assignment shapes gated on a literal-value
guard, as the issue asked.

- **Advisory, never blocking.** The warning attaches to whatever
  decision was already reached — `allow`, `ask`, `deny`, and the silent
  `unknown` fallthrough alike — and never changes it. On the `unknown`
  path the response carries no `permissionDecision`, so Claude Code's
  default prompt is untouched and the warning costs the fallthrough
  nothing.
- **Suggests the remediation, does not just refuse.** Emitted as
  `systemMessage` (the user sees it) *and* `additionalContext` (Claude
  sees it), which is what makes the fix actionable rather than
  decorative:
  ```
  KEY="$(fetch-secret)" sh -c 'curl -H "X-Api-Key: $KEY" https://service/endpoint'
  ```
- **Never logs or prints the matched value.** Shape and character offset
  only — a warning that quotes the secret puts the secret into the
  transcript it is warning about.
- **`YOLT_SECRET_WARN=0`** turns it off. A control that false-positives
  on legitimate work gets disabled wholesale and then protects nothing;
  a documented switch keeps that decision cheap and visible.

## Counting, for the "tighten later" step

No new log field needed — #86 already writes `[REDACTED:<shape>]` into
`~/.claude/yolt.log`, so the hit rate is greppable from the existing log
and can be weighed against real usage before anything is escalated
beyond advisory.

## Tests

`tests/test_secret_advisory.py` — 7 tests driving the real hook entry
point as a subprocess: warns on a *safe* `curl` without blocking, keeps
the safety verdict on an unsafe one, names shape and offset, includes
the env remediation, never leaks the value into the response JSON, stays
silent on clean commands and on `$GITHUB_TOKEN`-style expansions, and
honours the kill switch.

Full suite: 603 -> 610 tests, all green (3.12 and 3.14 locally, 3.10–3.12
in CI — which for this PR runs after retargeting to `master`, since the
workflow only triggers on PRs based on `master`).

## Version

`0.3.0` -> `0.4.0`. Minor: new user-visible warning and a new env var.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "yolt",
   "description": "Static safety analyzer for Bash invocations - auto-allows read-only commands and prompts for mutating ones, plus a Python AST analyzer for python3 / python3 -c / heredoc invocations.",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "author": {
     "name": "voitta-ai"
   },
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@
 - [Custom rules](#custom-rules)
   - [Python rules — `~/.claude/yolt/rules.json`](#python-rules---claudeyoltrulesjson)
   - [Shell rules — `~/.claude/yolt/shell.json`](#shell-rules---claudeyoltshelljson)
+- [Credentials in the command line (advisory)](#credentials-in-the-command-line-advisory)
 - [Debug / dogfood log](#debug--dogfood-log)
   - [Credential redaction](#credential-redaction)
 - [Self-improvement loop](#self-improvement-loop)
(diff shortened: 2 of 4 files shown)


debedb:
Superseded by #88. This PR was auto-closed by GitHub when its base branch (`feature/84-redact-secrets-in-logs`) was deleted on merging #86, and a closed PR whose base branch no longer exists cannot be reopened or retargeted. Same head branch, same commits, rebased onto master — no work lost.

Issue #75, opened 2026-07-26
Test suite is not hermetic: reads the developer's real ~/.claude/settings.json


debedb:
The test suite reads the ambient Claude Code settings of whoever runs it, so local results diverge from CI.

`hooks/yolt_analyzer.py:1098-1099` resolves allow-patterns from:

```
Path.home() / ".claude" / "settings.json",
cwd / ".claude" / "settings.json",
```

Those paths are live during the test run. Any `Bash(...)` allow pattern in the developer's real config short-circuits the classifier under test, turning expected `ask` decisions into `allow`.

Observed on master @ 1594ddd:

| environment | result |
|---|---|
| plain `python3 -m pytest tests/` | 19 failed, 556 passed |
| fake `$HOME` only | 5 failed, 570 passed |
| clean clone, no `.claude/`, fake `$HOME` | **575 passed** |
| GitHub Actions CI | green |

The 19 phantom failures split as: 14 from `~/.claude/settings.json` (109 `Bash(...)` allows) and 5 from an untracked local `.claude/settings.local.json` (50 more, including `Bash(git push *)` and `Bash(gh pr *)`).

CI is green only because the runner happens to have neither file. The suite is accidentally hermetic, not deliberately hermetic.

Impact: anyone running the suite locally sees ~19 failures unrelated to their change, and may "fix" non-bugs or dismiss a real regression as ambient noise.

Suggest: pin the settings-path resolution during tests — a fixture that points `Path.home()` and the cwd settings lookup at a temp dir, or an explicit `YOLT_SETTINGS_PATHS` env override the tests set. Same test/production isolation class as #72.

Issue #102, opened 2026-08-16
Umbrella: auto-mode realignment (v2.0.0)


debedb:
Tracking issue for the v2.0.0 realignment. Umbrella branch: **`feature/auto-mode-realignment`** (created off master at `aecdab6`).

## Why

Claude Code now ships **auto mode** — an LLM permission classifier on every tool call — as the *default* permission mode. That obsoletes the half of YOLT that granted approvals, and leaves untouched the half that detects credentials.

YOLT is being narrowed to a **deny-only guard** over a short list of decisions we refuse to delegate to a probabilistic classifier. It will stop emitting `allow` entirely, and it will cover every tool rather than only `Bash`.

## Structure

Per the `multi-phase-feature-pr-worktrees` skill: one umbrella branch off master, one `phase-N-*` branch per phase, **each phase PR targets the umbrella, not master**, each phase in its own worktree under `voitta-yolt.worktrees/`. The umbrella merges to master once all phases are in.

| Phase | Issue | Branch |
| --- | --- | --- |
| Ahead of the stack — README positioning, straight to master | #101 | own branch off master |
| 1 — stop emitting `allow` | #98 | `phase-1-no-allow` |
| 2 — widen past `Bash` to every tool (closes #80) | #99 | `phase-2-all-tools` |
| 3 — non-delegable list; delete the rules and the review loop | #100 | `phase-3-shrink` |
| 4 — deny-side git predicates from the closed #82 | #97 | `phase-4-git-predicates` |

## The gate — this is not optional

**Phase 1 is also the experiment. Phase 3 does not start until Phase 1 has run a full working day and the result is recorded on #98.**

Phase 1 is ~10 lines and trivially revertible. Phase 3 deletes hundreds. Running the cheap one first and *reading the result* is the entire reason for this ordering:

- prompts do not materially increase -> auto mode absorbs it, thesis confirmed, Phase 3 is safe;
- friction spikes -> the non-delegable list is bigger than we think, or auto mode is not sufficient — learned **before** anything is deleted.

Phase 2 and Phase 4 can proceed in parallel with the gate; neither is irreversible.

## Versioning and tagging

Phase PRs merge into the umbrella and **do not ship**, so they carry no version bump — CLAUDE.md's bump-every-shipping-merge rule is satisfied by the umbrella -> master PR carrying a single **1.0.0 -> 2.0.0**.

[...]

Issue #100, opened 2026-08-16
Phase 3: write the non-delegable list, delete the rest of the rules and the review loop


debedb:
Phase 3 of the auto-mode realignment stack. Targets `feature/auto-mode-realignment`, **not** master.

**Blocked on Phase 1's gate.** Do not start until Phase 1 has run a full working day and its result is recorded. This phase is the irreversible one.

## What

1. Write the non-delegable list.
2. Delete every rule that is not on it.
3. Delete the self-improvement loop.

## 1. The non-delegable list

This is the actual product decision and it does not exist yet. A deterministic guard earns its place only by encoding decisions we **refuse** to hand a probabilistic classifier — where "the model usually gets this right" is not an acceptable answer.

The list is expected to be short. Known candidates:

- writes to files that steer the agent (settings, hooks, skills, commands, agents, memory, MCP config) — the self-modification shape;
- command lines carrying a credential (already built, see `hooks/secret_redact.py`);
- the git-state predicates in #97;
- a small set of genuinely irreversible operations.

Everything else goes to auto mode, which does it better and free.

## 2. Delete the rest of the rules

`rules/shell.json` shrinks from "classify everything" to the list above. The per-CLI subcommand tables go. This is the treadmill that produced #52 #53 #66 #68 #69 #70 #73, all now closed as retired.

The retired content is not simply lost — biz#36 proposes translating the accumulated judgement into prose rules for the platform's own `autoMode.{allow, soft_deny, hard_deny}` block, where the compile target is natural language rather than a matcher. Capture it before deleting.

## 3. Delete the self-improvement loop

`hooks/yolt_review.py`, `commands/review.md` (`/yolt:review`), the SessionEnd hook registration, and the `~/.claude/yolt/` state it maintains.

Two reasons, and the second is the stronger one:

- It mines decision-log friction to suggest new rules. It is the treadmill's engine, and the treadmill is being retired.
- **It was independently the source of two credential leaks** (#91, #94). It copied raw log commands into `review.md` and `suggestions.json`, then — after the first fix — into `glob_collisions` as well, so the same credential appeared redacted on one line and in cleartext eight lines below. Deleting it removes two sinks from the blast radius permanently, which no amount of further redaction can match.

## Watch out

`secret_redact.py` must survive intact. It is the asset, and it is imported by the log writers, not by the reviewer. Verify the import graph before deleting anything.

## Version

None — merges into the umbrella.

debedb:
[...]

Issue #99, opened 2026-08-16
Phase 2: widen the hook past Bash to every tool


debedb:
Phase 2 of the auto-mode realignment stack. Targets `feature/auto-mode-realignment`, **not** master. Depends on Phase 1.

## What

Widen the hook from `Bash` to every tool, and fix the two gaps in #80 while doing it.

Auto mode covers every tool call. A guard that only covers `Bash` is not a guard — it is a guard on one door of a building with several.

## Scope

**Matcher.** `hooks/hooks.json` PreToolUse matcher `Bash` -> `*`.

**The hard part: everything downstream assumes argv.** `grammar_classifier.py` and `rule_classifier.py` take a shell command string and reconstruct argv. `Write`, `Edit`, `WebFetch` and MCP tools have structured input and no argv at all. This phase must decide what the gate even evaluates for them, and the answer should be small:

- the credential advisory, which is content-based and already tool-agnostic in principle;
- a path-based deny set for writes to agent-steering files (settings, hooks, skills, memory, MCP config) — the same category `safeclaude` classifies at the filesystem layer;
- nothing else. This is not an invitation to build a second classifier for structured tools.

**#80 gap 1 — the subagent deny ignores `permission_mode`.** `yolt_analyzer.py` gates on `agent_id` alone, so a subagent running under `bypassPermissions`, `dontAsk` or `auto` gets denied even though it would never have prompted and there was no hang to prevent. Either the deny is enforced (a regression — the operator explicitly pre-authorized and YOLT overrides them) or it is ignored (harmless, but the code claims a protection it does not deliver). Hooks fire in every mode and should respect the mode hierarchy.

**#80 gap 2 — subagent deny covers only Bash.** Resolved by the matcher change above.

## Carry Phase 1's rule forward

Widening the matcher must not reintroduce a grant. Non-Bash tools get `deny` / `ask` / silent fallthrough, same as Bash.

## Closes

#80.

## Version

None — merges into the umbrella.

debedb:
## Merged into the umbrella as Phase 2 — deliberately still open

#106 merged as `39bb1da` into `feature/auto-mode-realignment`. **This issue
and #80 stay open until the umbrella lands on master**, same as #97 and #98:
the code is on the umbrella, it has not shipped, and closing now would say
otherwise. Closing keywords cannot fire on a non-default-branch merge in any
case.

## What landed

Matcher `Bash` -> `*`, plus the small structured-tool answer this issue asked
[...]

PR #104, opened 2026-08-16
Phase 1: stop emitting `allow` — YOLT withholds approval, never grants it

### LINKED ISSUE #98: Phase 1: stop emitting `allow` (and measure what happens)
Phase 1 of the auto-mode realignment stack. Targets `feature/auto-mode-realignment`, **not** master.

## What

Stop emitting `permissionDecision: allow`. The hook returns `deny`, `ask`, or falls through silently to `unknown`, and nothing else.

## Why

Claude Code now ships auto mode — an LLM permission classifier on every tool call — as the **default** permission mode. Auto mode deliberately ignores broad pre-approvals in `permissions.allow`, because a blanket pre-approval defeats per-action vetting. `/auto-mode-setup` goes further and offers to **delete** such entries, flagging them as "classifier-bypassing".

A PreToolUse hook returning `allow` is the identical bypass, and that audit cannot see it. Local logs show **794 approvals granted this way while auto mode was active**. We built a hole in the thing we were reinforcing, and we are the only ones who can see it.

## Scope

- `hooks/yolt_analyzer.py` — the `safe` -> `allow` mapping goes. `safe` becomes a silent fallthrough, indistinguishable to Claude Code from `unknown`.
- Keep the decision in the log. `safe` stays a *classification*; it just stops being a *grant*. The log is how Phase 1's gate below gets measured.
- **Likely dead as a consequence:** the allow-pattern resolution added in `e610f3e` ("Honor explicit Bash allow patterns"). If nothing emits `allow`, reading the user's `permissions.allow` to short-circuit has no consumer. Verify rather than assume — and if it does go, it plausibly closes #75 (test suite reads the developer's real `~/.claude/settings.json`) as a side effect, since that is the code doing the reading.
- Tests asserting `allow` need re-pointing, not deleting: the assertion becomes "no `permissionDecision` in the response".

## This phase is also the experiment — gate the rest of the stack on it

**Do not start Phase 3 until this has run for a full working day.**

The whole thesis is that auto mode absorbs the friction this was removing. Phase 1 tests that for real, and it is trivially revertible.

- **Thesis confirmed** — prompts do not materially increase; auto mode picks it up invisibly. Deleting 700 lines of rules in Phase 3 is then safe.
- **Thesis wrong** — friction spikes. Then the non-delegable list needs to be larger than we think, or auto mode is not sufficient, and we learn it **before** deleting anything.

Measure from `~/.claude/yolt.log` (decision counts by type and by `permission_mode`) plus the subjective read. Record the result on this issue either way.

## Not in scope

Widening past `Bash` (Phase 2), shrinking the rules (Phase 3), deny-side predicates (#97).

## Version

None. Phase PRs merge into the umbrella branch and do not ship; the umbrella -> master PR carries the single 2.0.0 bump.

debedb:
Phase 1 of #102. **Targets `feature/auto-mode-realignment`, not master.** Closes #98, and #75 as a verified side effect.

## What

Two grant paths existed. Both are gone.

**1. `safe` -> `permissionDecision: allow`.** Now a silent exit, identical to `unknown`. The classification is still written to the decision log — that is what Phase 1's measurement reads — it just stops being an answer given on the host's behalf.

**2. The user-allow-pattern upgrade.** `GrammarClassifier` read `permissions.allow` from `~/.claude/settings.json` and the project settings and promoted any matching `unknown` or `unsafe` node to `safe`. That is sourced from your config rather than YOLT's rules, but it is a grant all the same — and after change 1 its only remaining effect would have been to **mislabel the log, corrupting the very measurement this phase exists to take.**

`load_allow_patterns` / `match_allow_patterns` stay in `rule_classifier` with their unit tests; only the wiring is removed. Deleting the rest is Phase 3 cleanup.

## Why

Auto mode deliberately ignores broad pre-approvals in `permissions.allow`, and `/auto-mode-setup` offers to delete them as "classifier-bypassing". A hook returning `allow` is the identical bypass and that audit cannot see it. The local decision log showed **794 approvals granted this way while auto mode was active**.

## #75 confirmed, not assumed

#98 predicted this would fall out. It did. The suite was non-hermetic because the hook resolved allow patterns from the developer's real `~/.claude/settings.json` *during the test run*, so a local `Bash(...)` entry turned an expected `ask` into an `allow`. Nothing reads those files now:

| | result |
|---|---|
| fake `$HOME` | **659 passed** |
| real `$HOME` (366 allow entries) | **659 passed** — previously ~19 phantom failures |

## Tests were re-pointed, not deleted

Worth reading, because two of these are places where the obvious edit would have made the suite weaker:

- Safe-path hook tests assert **no `permissionDecision` at all** rather than `allow`.
- `TestClassifierAllowPatterns` and `TestHookWhitelistDiscovery` tested the removed upgrade. Replaced with the **inverse** assertion — a matching `permissions.allow` entry must *not* change a decision — so the removal is pinned rather than merely untested.
- The adversarial catalogue's safe-side repros asserted hook `allow` as a proxy for "classified safe". Mechanically that becomes `assertIsNone` — but an `unknown` satisfies that too, so the repro would silently stop testing what it was written for. They now assert the classifier decision directly via a new `classification_for` probe, which is **stricter** than what it replaced.
- `test_hook_survives_and_still_classifies` and `test_unwritable_log_path_does_not_break_hook` used a *safe* command to prove the hook still produced output. A safe probe can no longer tell a healthy hook from a broken one, so they use a mutating probe and the log record respectively.
- The one deny-path escape hatch is pinned closed: `echo x > /etc/profile` can no longer be talked down to safe by any user setting.

## README

Decision mapping, the whitelist section, the intro line and the superseded use case are corrected, and the #103 warning block is rewritten now that the bypass is actually gone. The deeper rules documentation is untouched — Phase 3 deletes it.

## Version

None. Phase PRs merge into the umbrella and do not ship; the umbrella -> master PR carries the single `1.0.1` -> `2.0.0`.

## Next

**This PR is the gate.** Once it is on the umbrella, run a working day against it and record the result on #98 before Phase 3 (#100) starts. Phase 2 (#99) and Phase 4 (#97) are not blocked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
   - [Releasing (maintainers)](#releasing-maintainers)
   - [Migrating from manual to plugin install](#migrating-from-manual-to-plugin-install)
   - [Manual install (without the plugin system)](#manual-install-without-the-plugin-system)
-- [User whitelist as a secondary upgrade pass](#user-whitelist-as-a-secondary-upgrade-pass)
+- [User whitelist (removed in the auto-mode realignment)](#user-whitelist-removed-in-the-auto-mode-realignment)
 - [Dependencies](#dependencies)
 - [What the grammar classifier handles](#what-the-grammar-classifier-handles)
 - [SQL CLIs](#sql-clis)
@@ -47,8 +47,8 @@ and the platform now does it for every tool, with no rules to maintain.
 being narrowed to two things the platform does not do:
 
 - a **deny-only** guard over a short list of decisions that should not be
-  delegated to a probabilistic classifier — it will withhold approval and
-  never grant it;
+  delegated to a probabilistic classifier — it withholds approval and
+  never grants it;
 - **credential hygiene** — catching secrets on a command line before they
   are scattered across transcripts, logs and history, and keeping them out
   of YOLT's own records. See
@@ -59,23 +59,25 @@ The realignment is tracked in
 [#102](https://github.com/voitta-ai/voitta-yolt/issues/102) and lands in
 v2.0.0.
 
-> **If you run auto mode today, know this.** This version still returns
-> `allow` for commands it classifies as safe, and Claude Code honors a
-> hook's decision. So on those commands YOLT answers *instead of* auto
-> mode's classifier — the same bypass that `/auto-mode-setup` looks for in
-> your `permissions.allow` and offers to remove, except a hook is invisible
-> to that check. Removing it is the first change in v2.0.0
-> ([#98](https://github.com/voitta-ai/voitta-yolt/issues/98)). Until then,
-> if you want every command vetted by auto mode, uninstall the plugin.
+> **The classifier bypass is gone.** Up to v1.0.1 YOLT returned `allow`
+> for commands it classified as safe, and Claude Code honors a hook's
+> decision — so on those commands YOLT answered *instead of* auto mode's
+> classifier. That is the same bypass `/auto-mode-setup` looks for in your
+> `permissions.allow` and offers to remove, except a hook is invisible to
+> that check. As of
+> [#98](https://github.com/voitta-ai/voitta-yolt/issues/98) YOLT never
+> emits `allow` from any path, and no longer reads your settings files at
+> all. Every command it does not object to is vetted by auto mode.
 
-Everything below this section describes the tool as it is **today**, not as
-it will be after v2.0.0.
+The remaining phases shrink the rule set and widen the guard past `Bash`;
+until they land, the rules documentation below still describes the full
+classifier.
 
 ## Introduction
 
 A Claude Code hook that statically analyzes script invocations before
-execution, auto-allowing read-only ones and flagging mutating ones for
-review.
+execution and flags mutating ones for review. It withholds approval; it
+never grants it.
 
 > **Superseded framing.** The two gaps below were real when the built-in
 > whitelist matcher was the only thing standing between an agent and a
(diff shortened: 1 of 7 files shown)


debedb:
## Adversarial review

Codex was the intended independent reviewer and is **quota-blocked** — `codex-companion adversarial-review` returns `You've hit your usage limit ... try again at Sep 13th, 2026`. So this is a self-review, which is weaker by construction: I wrote the code and share its blind spots. Flagging that rather than presenting it as independent. (Question about our Codex entitlement raised separately.)

Method borrowed from the redactor rounds: take each new test, run it against the **pre-change** code, and require it to fail. A test that passes on both sides pins nothing.

### 1. Two of my own pins were vacuous — fixed in 1749aa9

`TestAllowHintSuggestions.test_allow_pattern_no_longer_upgrades_a_decision` and `test_protected_redirect_is_not_overridable` both **passed against the old code**. They asserted that an unknown stays unknown and a protected write stays unsafe — true before the change too, because `_make_classifier()` supplied no patterns, so the upgrade path they claimed to pin was never exercised.

Worse than useless: a future reader would have trusted them as the guard against reintroducing the grant.

Replaced with `test_classifier_refuses_allow_patterns_outright`, which asserts the constructor and `classify_command` now raise `TypeError` on an `allow_patterns=` kwarg. Verified in both directions — fails on old, passes here. The redirect test is renamed and its comment now states plainly that it does *not* pin the removal, naming the two that do.

The end-to-end pin was already sound: `TestHookIgnoresUserAllowPatterns` writes a real settings file and **2 of its 3 cases fail against the old code**.

### 2. Behavioural finding — real friction regression, and a confound for the gate

This is the one that matters and it is not fixed, because it may be correct.

The hook runs **even when a `permissions.allow` entry matches** (confirmed empirically — the local log contains commands covered by the user's own allow patterns). Previously the allow-pattern upgrade meant YOLT returned `allow` and stayed out of the way. Now YOLT returns `ask`, and the operator's explicit allow rule **cannot save it**, because a hook decision is authoritative.

Measured against the real log (106 user allow patterns, 8,566 decisions):

| decision | also matched a user allow pattern | effect after #98 |
|---|---|---|
| safe | 1350 (26.0%) | none — was double-covered, host's own rule still allows |
| unknown | 74 (5.0%) | none — falls through, host's rule allows |
| **unsafe** | **82 (4.3%)** | **these become new prompts** |

Concentrated in `find*`, `cat*`, `gh api*`, `gh pr merge*`, `git push origin feature/*`.

**Why it matters for the gate:** on the measurement day, some fraction of observed friction will be *YOLT overriding the operator's own allow rules*, which is a different failure than *auto mode did not absorb it*. Reading them as one number would give the wrong answer in either direction. #98's criteria updated to separate them.

Whether it should be fixed is a #100 question, not a #104 one: if something is on the non-delegable list, a user allow rule arguably **should not** silence it — that is what non-delegable means. The alternative (stay silent rather than ask when a user pattern matches) would preserve "never grant" but reintroduce the settings read and #75 with it.

### 3. Clean

- No remaining emitter of `allow` anywhere in `hooks/`.
- No dangling `seg` / `match_string` from the deleted call sites.
- **Nothing under `hooks/` reads settings files any more** — verified, so the README's "no longer reads your settings files at all" is accurate rather than aspirational. `yolt_review.py` only *mentions* settings.json in generated advice text; it does not read it.
- The credential advisory still fires on the safe path, which lost its old carrier response. Verified: no `permissionDecision`, `systemMessage` and `additionalContext` both present, token not echoed.

### Not covered by this review

An independent model. When Codex quota returns, worth re-running before the umbrella merges to master.

Issue #97, opened 2026-08-15
Reuse #82's git-state predicates deny-side: first entries in the non-delegable list


debedb:
Successor to #82. That PR built three deterministic git-state predicates and
wired them to **grant** approval. The predicates are good; the direction is
wrong. This issue keeps them and inverts them.

## Context

Claude Code now ships auto mode — an LLM permission classifier on every tool
call — as the **default** permission mode. YOLT is being narrowed to a
deny-only guard over a short list of decisions we refuse to delegate to a
probabilistic classifier. It will stop emitting `allow` entirely.

That list does not exist yet. **This issue proposes its first concrete
entries**, and they come from #82.

## What #82 established that is worth keeping

Its opening diagnosis is correct and survives the strategy change:

> *"does this command mutate anything?"* has no good answer for `git commit`.

Committing to a shared branch and committing to a worktree branch you opened
five minutes ago are different acts with **identical argv**. No purely
syntactic classifier can separate them. Answering it requires repository
state.

Three predicates from `hooks/git_policy.py`, all deterministic and
reproducible:

- **`linked_worktree`** — `--git-dir` != `--git-common-dir`; this is not the
  primary checkout.
- **`non_default_branch`** — default resolved from `origin/HEAD`, falling
  back to whichever of `init.defaultBranch` / `main` / `master` actually
  resolves *in this repo*. `init.defaultBranch` is a global preference for
  repos that do not exist yet, so it is a candidate and never an answer.
  Detached HEAD never qualifies.
- **`solo_author`** — every commit in `<default>..HEAD` was authored **and**
  committed by `git config user.email`. A branch with no commits of its own
  is vacuously solo.

Plus the supporting machinery: directory tracking through literal `cd` in
source order, `git -C <path>` retargeting, per-directory memoization, and
schema validation that rejects unknown predicate names at hook load.

## Proposed deny-side entries

Same probes, opposite conclusion. Candidates for the non-delegable list:

| Guard | Predicate |
| --- | --- |
| Block `git push` to the default branch | `deny_default_branch_refspec` — already implemented in #82's schema |
| Block `git push --force` / `--delete` against a branch carrying commits not authored by you | `!solo_author` |
| Block `git commit` / `git push` from the primary checkout where the repo convention is worktree-per-branch | `!linked_worktree`, config-gated — this is a house convention, not a universal truth |

These are the shape #34 asks for: crisp factual predicates, deterministic
[...]

PR #106, opened 2026-08-16
Phase 2: widen PreToolUse past Bash to every tool

### LINKED ISSUE #80: Subagent deny ignores permission_mode, and covers only Bash — scope + guard follow-up to #77
Follow-up to #76 / #77 (merged `efd6002`). Two gaps, plus the empirical confirmation that #77's premise was right.

## First: #77's premise is now confirmed, not inferred

#76 asserted that a gated call from a background subagent hangs with no prompt and no timeout. Nothing had actually observed it — skillz#127's log forensics showed those agents never reached the hook. A controlled probe (voitta-ai/skillz#137) settled it. Probe agent transcript, in full:

```
21:02:19  user        (the brief)
21:02:21  assistant   text: "I'll follow these steps exactly in order."
21:02:22  assistant   tool_use: Write -> /private/tmp/.../probe.log
<no tool_result, ever>
```

~2h37m, no result, no error, no timeout, no prompt to the operator. So `ask` from a subagent really is a hang, and `deny` really is the better answer. #77 was right.

## Gap 1: the deny ignores `permission_mode`

`hooks/yolt_analyzer.py:1147`:

```python
if agent_id:
    response = make_hook_response("deny", SUBAGENT_DENY_PREFIX + message)
```

Gated on `agent_id` alone. A subagent running under `bypassPermissions`, `dontAsk`, or `auto` would never have prompted, so there was no hang to prevent — but it now gets denied. Either the deny is enforced (a regression: the operator explicitly pre-authorized, and YOLT overrides them) or it is ignored in those modes (harmless, but the code claims a protection it does not deliver). The [hooks docs](https://code.claude.com/docs/en/hooks) say hooks fire in every mode and should respect the mode hierarchy.

Fix is one condition, using the field #77 already reads: deny only in modes that would actually have prompted.

## Gap 2: scope — the wedge is not a Bash problem

The probe hung on **`Write`**, not Bash. YOLT's matcher is `Bash`, so it was never in the path — same as skillz#127, where the six wedged agents never issued a Bash call at all. Across both observed incidents, YOLT's deny would have fired in neither.

That does not make #77 wrong; it makes it a partial mitigation of a broader class. Worth stating in the README next to the decision mapping, so "YOLT denies in subagents" is not read as "subagent wedges are handled". The general fix belongs upstream in Claude Code (a subagent permission request should surface or time out, not hang) — being reported separately.

## Unrelated, observed while investigating

#77's test run wrote ~314 fixture records into the real `~/.claude/yolt.log` — `agent_abc123`, `/tmp/wt-76`, `rm -rf /tmp/foo`, `somecommand_unknown_xyz --flag`. Live evidence for #72 (route fixture runs to a separate log) and #75 (tests are not hermetic): the suite writes to the production log, which is also the forensic artifact people reason from.

debedb:
Phase 2 of #102. **Targets `feature/auto-mode-realignment`, not master.** Closes #80.

## Why widen — the empirical version

Not symmetry with auto mode. **The subagent wedge in #80 hung on a `Write`**, so YOLT was never in its path; same in skillz#127, where six wedged agents never issued a Bash call at all. Across both observed incidents YOLT's deny would have fired in **neither**. A guard registered on one tool is not a guard.

## What the non-Bash path does — and what it deliberately does not

A structured tool has no argv, no shell, no rule lookup. Two things happen:

1. **Agent-steering write check** — a write to a path in `unsafe_write_targets` is `unsafe`. The self-modification shape, and the one thing on the non-delegable list a structured tool can reach today.
2. **Credential advisory** over every string in the payload, so a token in a `Write` body or nested in an MCP argument is caught like one on a command line. The wording drops the `argv` / `ps` framing — neither applies, and advice that does not apply is how a control earns its way into being ignored.

Everything else is `unknown`; the host decides.

`TOOL_WRITE_PATH_FIELDS` is a **closed list** (`Write`, `Edit`, `MultiEdit`, `NotebookEdit`). Guessing which field of an arbitrary MCP payload is a write target eventually guesses wrong, and a wrong guess is a false `deny` inside a subagent — the exact failure this exists to prevent. Four tests pin the closure, including an `mcp__fs__write_file` payload with a `path` argument pointing at a protected file, which correctly falls through.

**This must not become a second classifier for structured tools.** The host's own vetting covers the general case; a per-tool rule set is the treadmill the realignment is ending.

## #80 gap 1 — with a deliberate deviation

The `ask` -> `deny` conversion was gated on `agent_id` alone, so it fired in modes where the operator had blanket-authorised and no prompt could have hung. Now skipped under `bypassPermissions` and `dontAsk`.

**`auto` is not exempt, which departs from the mode list in the issue.** Auto mode delegates the *decision* to a classifier; it does not put an operator behind a hook's `ask`, so a subagent receiving one still has nobody to answer it. The costs are asymmetric — a wrong deny fails in seconds with a reason the agent reports upward; the wrong ask wedged an agent for **2h37m** in the #80 probe. Absent `permission_mode` denies for the same reason. Flagging it rather than quietly following the issue.

## Tests — verified non-vacuous

24 new tests. Run against the **pre-change** code, **16 of 24 fail**. The 8 that pass are fall-through guards (unlisted tool stays silent, ordinary path stays silent) which correctly hold on both sides — they pin the boundary, not the change. Saying so explicitly because the #104 review found me claiming pins that were not.

Also pinned: the advisory does not alter the decision (same call with and without a credential decides identically, with a vacuity guard requiring the credential case to actually warn), never echoes the value, and the `YOLT_SECRET_WARN` kill switch reaches the new path.

**683 tests, green.**

## Logging

Records gain `tool_name`. For non-Bash the `command` field carries the target path, not the payload — a `Write` body can be a megabyte and this log is append-only.

## README

Documents the new surface, and states the thing #80 asked for explicitly: **"YOLT denies in subagents" is not "subagent wedges are handled"**. The deny only converts YOLT's own ask; any other gated call in a background subagent still hangs, and that fix belongs upstream in Claude Code.

## Version

None — phase PRs merge into the umbrella and do not ship.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 - [Introduction](#introduction)
 - [Example use cases](#example-use-cases)
 - [How it works](#how-it-works)
+- [Tools other than Bash](#tools-other-than-bash)
 - [Install](#install)
   - [Updating](#updating)
   - [Releasing (maintainers)](#releasing-maintainers)
@@ -218,6 +219,51 @@ is safe but `python3 -m pip install` is unsafe); known CLIs use their
 `rules/shell.json` spec; wrappers (`time`, `xargs`, `timeout`, `env`,
 `nice`, `watch`, ...) re-classify the wrapped command; anything else →
 unknown.
+
+## Tools other than Bash
+
+The PreToolUse matcher is `*`, not `Bash`
+([#99](https://github.com/voitta-ai/voitta-yolt/issues/99)). The reason is
+empirical rather than tidy: the subagent wedge investigated in
+[#80](https://github.com/voitta-ai/voitta-yolt/issues/80) hung on a
+`Write`, so YOLT was never in its path. A guard registered on one tool is
+not a guard.
+
+A structured tool has no argv, no shell and no rule lookup, so exactly two
+things happen for it:
+
+1. **The agent-steering write check.** If the call writes to a path in
+   `rules/shell.json#unsafe_write_targets` — settings, hooks, skills,
+   commands, agents, memory, MCP config — it is `unsafe`. That is the
+   self-modification shape, and it is the one thing on the non-delegable
+   list a structured tool can reach today.
+2. **The credential advisory**, over every string in the payload, so a
+   token in a `Write` body or an MCP argument is caught the same as one on
+   a command line. The wording drops the `argv` / `ps` framing there,
+   because neither applies.
+
+Everything else is `unknown` and the host decides.
+
+The write-target field list is **closed** — `Write`, `Edit`, `MultiEdit`
+and `NotebookEdit`, by their documented path fields. Guessing which field
+of an arbitrary MCP payload is a write target eventually guesses wrong,
+and a wrong guess is a false `deny` inside a subagent, which is the exact
+failure this is meant to prevent. This is deliberately not a second
+classifier for structured tools: the host's own vetting covers the general
+case.
+
+> **“YOLT denies in subagents” is not “subagent wedges are handled.”**
+> The deny converts *YOLT's own* `ask` into a fast failure. Any other
+> gated call in a background subagent still hangs with no prompt and no
+> timeout — 2h37m with no result in the #80 probe. The general fix belongs
+> in Claude Code, where a subagent permission request should surface or
+> time out rather than hang.
+
+The conversion is skipped under `bypassPermissions` and `dontAsk`, where
+the operator blanket-authorised ahead of time and there was no prompt to
+hang on. It is **not** skipped under `auto`: auto mode delegates the
+decision to a classifier, it does not put an operator behind a hook's
+`ask`, so a subagent that receives one still has nobody to answer it.
 
 ## Install
 
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -2,7 +2,7 @@
   "hooks": {
     "PreToolUse": [
       {
-        "matcher": "Bash",
+        "matcher": "*",
         "hooks": [
           {
             "type": "command",
--- a/hooks/yolt_analyzer.py
+++ b/hooks/yolt_analyzer.py
@@ -873,6 +873,195 @@ SUBAGENT_DENY_PREFIX = (
     "the main session, or add a matching permissions.allow entry.\n\n"
 )
 
+# Permission modes in which the operator has blanket-authorised ahead of
+# time, so an `ask` from this hook is not a question anybody was going to
+# be asked. Converting it to `deny` inside a subagent would override an
+# explicit pre-authorisation rather than protect against a hang. Issue
+# #80, gap 1.
+#
+# `auto` is deliberately NOT in this set, which departs from the list in
+# #80. Auto mode delegates the *decision* to a classifier; it does not
+# put an operator behind a hook's `ask`, so a subagent that receives one
+# still has nobody to answer it. The costs are wildly asymmetric — a
+# wrong deny fails in seconds with a reason the agent can report upward,
+# a wrong ask wedged an agent for 2h37m in the #80 probe — so `auto`
+# keeps the deny.
+SUBAGENT_DENY_EXEMPT_MODES = frozenset({"bypassPermissions", "dontAsk"})
+
+# Tools whose input carries a filesystem path the tool writes to.
+#
+# Deliberately a closed list. Guessing which field of an arbitrary MCP
+# payload is a write target will eventually guess wrong, and the failure
+# mode of a wrong guess is a false `deny` in a subagent, which is exactly
+# the wedge this hook exists to avoid. An unlisted tool falls through to
+# `unknown` and the host decides. Issue #99.
+TOOL_WRITE_PATH_FIELDS = {
+    "Write": ("file_path",),
+    "Edit": ("file_path",),
+    "MultiEdit": ("file_path",),
+    "NotebookEdit": ("notebook_path",),
+}
+
+
+def tool_input_text(tool_input):
+    """Flatten a tool input to the text the credential scanner reads.
+
+    Every string anywhere in the payload, joined. A credential can ride
+    in a `Write` body, a `WebFetch` url, or an arbitrary MCP argument,
+    and enumerating the field that carries it per tool is the same
+    losing game as enumerating write-target fields — except here the
+    cost of over-reaching is a false warning rather than a false deny,
+    so this one reaches everywhere.
+    """
+    parts = []
+
+    def walk(value):
+        if isinstance(value, str):
+            parts.append(value)
+        elif isinstance(value, dict):
+            for item in value.values():
+                walk(item)
+        elif isinstance(value, (list, tuple)):
+            for item in value:
+                walk(item)
+
+    walk(tool_input)
+    retval = "\n".join(parts)
+    return retval
+
+
+def make_gated_response(message, permission_mode, agent_id):
+    """Build the response for a decision YOLT objects to.
+
+    `ask` normally, `deny` inside a subagent — no operator is reachable
+    from a background subagent, so the approval dialog never surfaces and
+    nothing times out; the agent wedges indefinitely (observed at 2h37m
+    in #80). `deny` is strictly more restrictive than `ask` and fails in
+    seconds with a reason the agent can report upward. Issues #76, #77.
+
+    The subagent conversion is skipped in the modes listed in
+    `SUBAGENT_DENY_EXEMPT_MODES`, where the operator pre-authorised and
+    there was no prompt to hang on. Issue #80, gap 1.
+    """
+    if agent_id and permission_mode not in SUBAGENT_DENY_EXEMPT_MODES:
+        retval = make_hook_response("deny", SUBAGENT_DENY_PREFIX + message)
+        return retval
+    retval = make_hook_response("ask", message)
+    return retval
+
+
+def _run_hook_non_bash(tool_name, tool_input, permission_mode, agent_id):
+    """PreToolUse path for every tool that is not Bash. Always exits.
+
+    The matcher covers `*` as of #99, because the wedge #80 observed was
+    a `Write`, not a Bash call — a guard registered on one tool is not a
+    guard. Two things happen here and nothing else:
+
+      1. the agent-steering write check (`classify_tool_input`), and
+      2. the credential advisory, over every string in the payload.
+
+    There is no shell parsing, no rule lookup and no argv: a structured
+    tool has none of those. Anything not covered by (1) falls through to
+    `unknown` so the host decides.
+    """
+    text = tool_input_text(tool_input)
+    advisory = format_secret_advisory(text, tool_name=tool_name or "tool")
+
+    try:
+        hooks_dir = Path(__file__).resolve().parent
+        if str(hooks_dir) not in sys.path:
+            sys.path.insert(0, str(hooks_dir))
+        from grammar_classifier import GrammarClassifier
+        from rule_classifier import DECISION_UNSAFE, load_shell_rules
+        yolt_dir = hooks_dir.parent
+        shell_rules = load_shell_rules(
+            rules_dir=yolt_dir / "rules",
+            user_overrides_path=Path.home() / ".claude" / "yolt" / "shell.json",
+        )
+    except Exception as e:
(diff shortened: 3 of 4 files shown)


debedb:
**Hold this merge until the #98 measurement window closes** — same reason I just applied to #97, and I should have applied it here first.

This PR adds a new deny surface: agent-steering writes via `Write` / `Edit` / `MultiEdit` / `NotebookEdit`. That is manufactured bucket-B friction. If it lands before the window, a `materially worse` verdict cannot be attributed — it could be auto mode failing to absorb Phase 1's delegation, or it could be this PR's new denies, and the measurement has no way to separate them.

I held #97 on exactly this argument and did not apply it to my own PR. Consistency: anything that adds a deny rule waits for the window.

Nothing wrong with the code — CI is green, the review is done. It merges the moment the window closes.

The dogfood build now running locally is the umbrella at `d6d5a01`, i.e. **Phase 1 only**, which is the correct isolation for what #98 is measuring.

debedb:
## Rebased onto the current umbrella and re-verified

This PR sat from 2026-08-16 while four security releases and two umbrella
syncs landed. Rebased onto `aee011d`; **803 tests pass**, up from the
umbrella's 779.

The rebase is worth more than currency here: this PR gates structured tools
by checking the write path against `_target_is_unsafe_write`, and that list
grew substantially while the PR was open. So Phase 2 inherits #127's
agent-steering entries and #128/#136/#138's normalisation **without knowing
anything about them**:

```
ASK      Write         ~/.claude/skills/evil/SKILL.md
ASK      Write         ~/.claude/settings.json
ASK      Write         ~/.ssh/authorized_keys
ASK      Edit          ~/.claude/agents/x.md
ASK      NotebookEdit  ~/.claude/skills/n.ipynb
SILENT   Write         /tmp/harmless.txt
SILENT   Read          ~/.claude/skills/evil/SKILL.md
```

`Write` to `~/.ssh/authorized_keys` asks, and Phase 2 contains no mention of
ssh — that is the shared list doing the work, which is the argument for one
list rather than a second classifier for structured tools.

`Read` on the same path stays silent: it is not in `TOOL_WRITE_PATH_FIELDS`,
so there is no write target to check.

## #80 gap 1, and a deliberate deviation from #99

Verified across the mode matrix, in and out of a subagent:

```
DENY   mode=auto               agent=yes
ASK    mode=auto               agent=no
DENY   mode=default            agent=yes
ASK    mode=bypassPermissions  agent=yes
ASK    mode=dontAsk            agent=yes
DENY   mode=acceptEdits        agent=yes
```

`SUBAGENT_DENY_EXEMPT_MODES` is `{bypassPermissions, dontAsk}`.

**#99 asks for `auto` to be exempt too. It should not be, and this
implementation is right to leave it out.** The deny exists because a subagent
has no operator to answer a prompt, so an `ask` wedges. Under
`bypassPermissions` and `dontAsk` nothing would have prompted, so the deny is
a pure regression — the operator pre-authorised and YOLT overrides them.
Under `auto` the host's own classifier still prompts, so the wedge is real
and the deny is doing its job.

Flagging rather than silently keeping it: the issue text is wrong on this
point and should be corrected, not the code.

## Scope held

Still the small answer #99 asked for — the credential advisory plus a
path-based write check, and nothing else. `TOOL_WRITE_PATH_FIELDS` is a
closed list, and its comment says why: guessing which field of an arbitrary
MCP payload is a write target eventually guesses wrong, and a wrong guess is
a false `deny` inside a subagent, which is the exact wedge this hook exists
to prevent. An unlisted tool falls through to the host.

## Version

None. Phase PRs merge into the umbrella; the umbrella -> master PR carries
the single 2.0.0.

debedb (commented):

Do not ship. Non‑Bash path widens critical surface without depth/size bounds, permits protected writes via unlisted tool/field variants, and risks leaking secrets in logs when the redactor is unavailable. These are high‑impact, hard‑to‑detect regressions on a hook that now runs for every tool.

### Next steps
- Add depth/size/cycle guards to tool_input_text and subject building; expand TOOL_WRITE_PATH_FIELDS to vetted MCP write tools; change logging to redaction-first-withhold and schema-bounded subjects; tighten non‑Bash mutate fallthrough semantics; add tests for cycles, bytes payloads, MCP writes, and missing-field mutate cases.

### Findings outside the diff
_(could not be attached inline; lines are not part of this PR's diff)_
- `hooks/yolt_analyzer.py:1560-1645` **[high] Unbounded recursive walk over tool_input risks stack overflow/wedge** _(confidence 0.74)_
  tool_input_text flattens dicts/lists by stringifying every scalar without recursion bound or cycle detection. If tool_input contains a self-referential structure (possible via hostile plugin/tool) or very deep nesting, this can recurse indefinitely or blow the call stack/CPU. hook runs on every non‑Bash tool; an infinite/very deep payload would wedge all tool calls or crash PreToolUse. Add explicit max-depth and seen-set (id-based) tracking; treat overflow as truncation rather than raising. _(fix: Implement tool_input_text with parameters: max_depth (e.g., 5–10), byte/char budget, and a seen=set() of object ids to break cycles. On overflow, return a placeholder and continue. Add tests for cyclic and >max_depth inputs.)_
- `hooks/yolt_analyzer.py:1330-1405` **[high] Decision log may capture secret-bearing payload slices** _(confidence 0.69)_
  _tool_log_subject for non‑Bash tools derives ‘command’ by coercing entire tool_input to a string and truncating. For Write edits this includes raw content bodies (megabytes or secret material) until string-based redaction runs. If secret_redact is unavailable (REDACTOR_IMPORT_ERROR set), _redact_or_withhold returns WITHHELD only for redact(None), but _tool_log_subject still feeds the pre-redaction subject into the 500‑char truncation path; tests show command contains ‘/tmp/ordinary.txt’ and omits ‘yyyy’, but a short credential value in the first 500 chars will be logged unless redactor is present. That leaks secrets into yolt.log despite “Not blocked” posture. _(fix: For non‑Bash, build subject from a bounded schema: include tool_name plus only path-like fields from TOOL_WRITE_PATH_FIELDS when present; never include arbitrary values (content, headers). If no safe summary, use a fixed token. Apply redaction first; when redactor is missing, default to WITHHELD rather than logging any slice of tool_input. Add a size cap before redaction to avoid quadratic cost.)_
- `hooks/yolt_analyzer.py:1140-1180` **[medium] Protected-path ask missing for MCP write tools** _(confidence 0.62)_
  TOOL_WRITE_PATH_FIELDS is declared as a closed map, but the MCP write surface isn’t represented beyond core tools. For example, tests assert mcp__fs__write_file falls through. That means writes to protected agent-steering paths via MCP methods bypass the ask-guard entirely if the host wires those tools (common in Codex/CLA). Impact: subagents can self-modify settings.json without prompt. The closed list is correct, but missing core MCP write tools is a coverage hole. _(fix: Extend TOOL_WRITE_PATH_FIELDS to include vetted MCP write tools that can target file paths (e.g., mcp__filesystem__write_file, mcp__filesystem__edit_file, mcp__filesystem__move_file with destination). Keep the list closed and add contract tests mirroring tests/test_non_bash_tools.py to verify protected-path gating for each added entry. If exact names vary, add a normalization layer with an allowlisted prefix+exact field mapping.)_

- `hooks/yolt_analyzer.py:1240-1315` **[high] Silent fallthrough on unlisted mutate tools/fields enables protected writes** _(confidence 0.58)_
  _run_hook_non_bash constructs responses via make_gated_response for protected writes, which produces ‘ask’ in main sessions and ‘deny’ in subagents except SUBAGENT_DENY_EXEMPT_MODES. However, classify_tool_input returns (“unknown”, “…no rule…”) for unlisted tools/fields, and _run_hook_non_bash then proceeds to credential-advisory and logging but never sets permissionDecision → the widened matcher silently lets all unlisted write-capable tools proceed, including rename/move/delete variants that mutate protected files via alternate fields. This undermines the stated goal to catch agent-steering writes across all tools. _(fix: For non‑Bash, if tool_name is known-mutate class (Write/Edit/MultiEdit/NotebookEdit) but field missing, fail closed to ‘ask’ with reason ‘unlisted write-target field’. Alternatively, add a minimal denylist for protected absolute targets detected anywhere in tool_input when tool_name is in the mutate class. Keep fallthrough only for clearly non‑mutating tools.)_
- `hooks/yolt_analyzer.py:1608-1668` **[medium] Bytes/opaque values can leak via subject stringification when redactor unavailable** _(confidence 0.51)_
  tool_input_text includes numbers and booleans by str() and ignores bytes-like values. If a tool passes bytes (e.g., file chunks, secrets) the str() on a Python bytes produces a repr beginning with ‘b’ which will still include raw content inside quotes; the credential scanner runs over this text, but _tool_log_subject then logs the same string-form before redaction. With redactor missing, small secrets inside bytes payloads leak into yolt.log. The advisory text promises not to echo values, but the decision log contradicts that under redactor-outage. _(fix: Before serialization, drop large/value fields by key (content, data, body, headers, args) from the subject builder; for tool_input_text, treat bytes as opaque: count length only and emit ‘<bytes:N>’. Keep redaction-first and WITHHELD fallback.)_

</details>

debedb:
Five findings. **One is real and serious — fixed in `4f6e782`. Two are
refuted by measurement. Two are the documented design, and worth arguing
about rather than patching.**

## Real: the widened matcher could exit non-zero (0.74)

```
depth 100     exit=0
depth 1000    exit=1   RecursionError: maximum recursion depth exceeded
depth 20000   exit=1
```

Every other failure path in `run_hook` exits 0 deliberately — a guard that
cannot classify should cost the host nothing. An uncaught exception was the
one way that promise could break, and **widening the matcher to `*` is what
made arbitrary payload shapes reachable**: the credential walk now runs on
every tool call, including MCP payloads this repo does not author.

Fixed in two layers, because the bound fixes the instance and the guard fixes
the class:

- the walk is bounded at 64 levels, and exceeding it **truncates rather than
  raises** — a missed credential warning is advisory, a raised exception
  costs the tool call;
- `run_cli`'s `--hook` dispatch catches everything that is not `SystemExit`
  and exits 0.

Mutation-checked, and the layering shows in the numbers: unbinding the walk
alone fails **one** test, because the boundary guard still catches the
`RecursionError`. Removing both fails **four**.

The cycle half of the finding does not apply — `tool_input` arrives as parsed
JSON, and JSON cannot express a self-referential structure. Depth was the
real half.

## Refuted: secrets reaching the decision log (0.69)

Measured with a real token in a `Write` body:

```
redactor available     command field: 'Write /tmp/ordinary.txt'
redactor UNAVAILABLE   command field: '[WITHHELD:redactor-unavailable]'
```

The secret appears in neither. `_tool_log_subject` builds a *subject* — tool
name plus path — not a stringified payload. The finding assumed it coerces
the whole `tool_input`; it does not.

## Refuted: bytes leaking through stringification (0.51)

```
json.dumps({"a": b"secret"}) -> TypeError: Object of type bytes is not JSON serializable
```

`tool_input` is parsed JSON. A bytes value cannot reach `tool_input_text` at
all.

## Design, not defect: MCP write tools absent from the closed list (0.62, 0.58)

Both findings say the same thing: a write to a protected path through an MCP
tool is not gated, because `TOOL_WRITE_PATH_FIELDS` is closed.

That is correct and deliberate, and the code says why at the definition:
guessing which field of an arbitrary MCP payload is a write target will
eventually guess wrong, and **the failure mode of a wrong guess is a false
`deny` inside a subagent** — which is precisely the wedge this hook exists to
avoid. An unlisted tool falls through to the host, which has its own
permission layer.

The finding's recommendation — extend the list to "vetted MCP write tools" —
is the right shape but cannot be done blind. `mcp__fs__write_file` is only an
example; the real set depends on which servers a given operator has wired,
and this repo cannot enumerate that.

**What I would accept as the actual fix** is configuration rather than a
shipped list: let the operator declare their MCP write tools and their path
field, the same way `rules/shell.json` already carries operator overrides.
That is a real gap, it is out of scope for Phase 2 under #99's explicit
"nothing else", and it deserves its own issue rather than a guess bolted on
here.

## On the run

Fourth degenerate payload of this session on the first attempt — and this
time self-aware about it: *"The previous message was a brief status note, not
the final adversarial review."* Re-running with the focus rewritten to
"your entire output must be findings, no preface, no next steps; if you have
no finding, emit nothing" produced the five above.

Two of five held up under test, and the one that did is a session-wedging
crash that only exists because of this PR's own matcher change. That is the
review earning its place.

PR #116, opened 2026-08-28
Keep the newlines when rebuilding a multi-line double-quoted argument

### LINKED ISSUE #115: Newlines are dropped from multi-line double-quoted arguments, so every multi-line python3 -c is unanalyzable
## What happens

Every `python3 -c "..."` whose script spans more than one line is classified
`unsafe` with the reason:

```
could not statically analyze inline python3 -c script (parser bailed at line 1)
```

It is always line 1, and it happens whether the script is read-only or not.

## Why

`GrammarClassifier._reconstruct_string` (`hooks/grammar_classifier.py`) rebuilds
a double-quoted string by concatenating the slices of the string node's
children. tree-sitter-bash ends a `string_content` run at each newline and
starts a fresh one after it, and the newline byte between them belongs to no
child at all:

```
string '"\nimport json,sys\nd=json.load(sys.stdin)\n..."'
  string_content 'import json,sys'
  string_content 'd=json.load(sys.stdin)'
  string_content "print(d.get('ok'))"
```

So the reconstruction welds each line onto the next. What reaches
`SafetyAnalyzer.analyze` is:

```
import json,sysd=json.load(sys.stdin)print(d.get('ok'))
```

`ast.parse` raises `SyntaxError: invalid syntax (line 1, offset 17)` -- offset
17 is `sysd`, the seam between the first two lines. `classify_python_source`
sees `parse_error` and, on the inline path, returns unsafe with the message
above. There is only ever a line 1 after the weld, which is why the line
number never varies.

The direction is safe -- fail-closed -- but the answer is wrong, and it is
wrong for a whole shape of command rather than an edge case.

## Why it went unnoticed

The multi-line arguments already under test are SQL
(`test_sqlite3_multiline_select_safe`), and SQL with a newline removed is
still valid SQL. Python is where a lost newline stops being cosmetic.

## Scope

Only double-quoted strings. A `raw_string` (`python3 -c '...'`) is taken as a
single slice, so single-quoted inline scripts are unaffected.

## Repro

```
python3 hooks/grammar_classifier.py 'python3 -c "
import json,sys
print(json.load(sys.stdin))
"'
```

## Downstream

Reported from voitta-ai/shmobster, which gates shell commands through this
classifier: read-only inspection scripts were being parked for human approval
in Slack.

debedb:
Closes #115.

## The bug

`GrammarClassifier._reconstruct_string` rebuilt a double-quoted string by
concatenating the slices of the string node's children. tree-sitter-bash ends
a `string_content` run at each newline and starts a fresh one after it, and
the newline byte belongs to no child -- so every line got welded onto the
next.

Where that mattered is inline Python. A multi-line `python3 -c "..."` reached
`SafetyAnalyzer.analyze` as a single line, `ast.parse` raised at the seam, and
`classify_python_source`'s inline path turned that into:

```
could not statically analyze inline python3 -c script (parser bailed at line 1)
```

Always line 1 -- after the weld there is no other line -- and for every
multi-line inline script, read-only ones included. Fail-closed, so the answer
was safe in direction and wrong in substance: read-only inspection scripts got
parked for human approval.

## The fix

Carry the bytes between adjacent children across, verbatim. They are literal
string text, and the analyzer downstream wants them exactly as the shell would
pass them -- normalizing to `\n` would be a guess. The gap fill runs before the
per-child dispatch rather than instead of it, so a command substitution inside
a multi-line string is still masked with the placeholder.

Only double-quoted strings were affected; a `raw_string` (`python3 -c '...'`)
is taken as one slice.

## Before / after

| command | before | after |
| --- | --- | --- |
| `curl … \| python3 -c "<multi-line read-only>"` | unsafe -- parser bailed at line 1 | `safe` -- `curl: read-only; python: python3 -c ...` |
| `cd … && python3 -c "<multi-line read-only>"` | unsafe -- parser bailed at line 1 | `safe` -- `builtin: cd; python: python3 -c ...` |
| `python3 -c "<multi-line shutil.rmtree>"` | unsafe | unsafe (now on the finding, not the parse error) |

## Tests

Five in a new `TestMultilineDoubleQuotedString`. Four fail on master; the
fifth is the guard that a destructive multi-line script stays unsafe once it
does parse. Two assert on the rebuilt argv rather than on a decision, since a
decision only proves the source parsed -- it would still pass if the newlines
came back as spaces.

Full run: 200/200 in `test_grammar_classifier`. The suite-wide run has
pre-existing failures in `test_yolt_hook` that reproduce on master and are
#75 (the tests read the developer's real `~/.claude/settings.json`).

## Version

`1.0.4` -> `1.1.0`. Minor: this flips a real class of command from unsafe to
safe, which is user-visible at the permission prompt.

## Where it was found

voitta-ai/shmobster gates shell commands through this classifier and posts an
approval card to Slack for anything not `safe`. Three cards in one thread, two
of them this bug.

https://claude.ai/code/session_01DrsoG6XaCr9VvWHfpqcMVX

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "yolt",
   "description": "Deterministic pre-execution guard for Bash: decomposes compound commands, heredocs and command substitutions with tree-sitter, analyzes inline Python by AST, flags credentials on the command line, and redacts them from its own logs. Narrowing to deny-only in v2.0.0 now that Claude Code auto mode vets tool calls.",
-  "version": "1.0.4",
+  "version": "1.1.0",
   "author": {
     "name": "voitta-ai"
   },
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "yolt",
-  "version": "1.0.4",
+  "version": "1.1.0",
   "description": "Deterministic pre-execution guard for Bash: decomposes compound commands, heredocs and command substitutions with tree-sitter, analyzes inline Python by AST, flags credentials on the command line, and redacts them from its own logs. Narrowing to deny-only in v2.0.0 now that Claude Code auto mode vets tool calls.",
   "author": {
     "name": "voitta-ai"
--- a/hooks/grammar_classifier.py
+++ b/hooks/grammar_classifier.py
@@ -289,8 +289,30 @@ class GrammarClassifier:
         return self._slice(node, src)
 
     def _reconstruct_string(self, node, src):
+        """Rebuild the value of a double-quoted string from its children.
+
+        The bytes between two adjacent children are part of the string and
+        have to be carried across (#115). tree-sitter-bash ends a
+        `string_content` run at a newline and starts a fresh one after it,
+        so a multi-line `"..."` arrives as one child per line with the
+        newlines belonging to no child at all. Concatenating only the
+        children welds every line onto the next: a multi-line
+        `python3 -c "..."` reached the Python analyzer as a single line,
+        `ast.parse` raised on it, and the inline path turned that into
+        "could not statically analyze inline python3 -c script (parser
+        bailed at line 1)" -- for every such command, since after the weld
+        there is only ever a line 1. The gate then parked read-only scripts
+        for a human, which is the safe direction but the wrong answer.
+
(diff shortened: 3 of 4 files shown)


debedb (commented):
## Codex Adversarial Review — approve

No ship-blocking issue found in the branch diff. I inspected the changed code, ran the targeted multiline tests and the full grammar classifier suite successfully, and both independent review lanes reported no substantive findings. Validation gap: LSP diagnostics were unavailable in this session.

Issue #119, opened 2026-08-28
The 500-char command cap in the decision log costs the Phase 3 gate 39% of its corpus


debedb:
## What

`_log_hook_decision` and `_log_ran_command` both store `command[:500]`
(`hooks/yolt_analyzer.py:1098`, `:1152`). On the current dogfood corpus that
truncates **2,459 of 6,305 unique commands — 39%**.

## Why it matters now

The Phase 3 gate (#100) replays the decision log through a candidate
classifier and diffs the `unsafe` set. A truncated record cannot be replayed
faithfully: the cut discards the command's trailing segments, and those are
disproportionately the mutating ones (`&& git push`, `&& gh pr merge`), so the
replay classifies a fragment that was never what the hook saw.

Measured with `scripts/replay_unsafe.py` against the build that produced the
log:

| subset | reproduces its logged verdict |
| --- | --- |
| intact (< 500 chars) | 99.6% (3830/3846) |
| truncated (== 500) | 54.9% (1339/2438) |

1,099 of the 1,115 total disagreements are truncation alone. The harness
excludes truncated records by default, so the gate currently runs on 61% of
the corpus it could have had.

## Why this is not just "raise the number"

**Redaction runs before truncation** — that ordering is deliberate and
documented at `yolt_analyzer.py:1080`. So the cap is also, incidentally, a
bound on how much command text a leaked-secret-in-a-log incident exposes.
Raising it enlarges the surface `secret_redact.py` has to get right, on a
file that has already been the subject of two credential-leak fixes (#91,
#94) and a pre-existing warning that historical log lines still hold
plaintext.

So this is a security tradeoff, not a constant to bump. Options, roughly in
increasing cost:

1. **Raise the cap and accept the surface**, e.g. to 2,000 chars. Cheapest,
   and the redactor is materially better than it was at the time of #91/#94.
2. **Store a hash of the full command alongside the truncated text.** Does not
   help replay -- you cannot classify a hash -- but it would at least let a
   future analysis detect that two truncated records were different commands,
   which today it cannot.
3. **Log the full command to a separate, mode-gated file** that only exists
   while a measurement window is running, and is not written by default.
   Keeps the steady-state exposure exactly where it is.

(3) is the one that actually serves the gate without widening the default
blast radius, but it is also the most work and adds a config surface.

## Not urgent

The gate runs today on 3,846 intact commands, which is a large sample. This
is filed so the 39% is a known, measured limitation rather than something
[...]

PR #120, opened 2026-08-28
Enumerate the asks Phase 3 would delete, and refuse to guess when the log cannot say

debedb:
Builds the gate that #100 was re-pointed at, and finds that the gate as I
specified it could not have run correctly.

## Phase 3 does not need a measurement window

Post-Phase-1 the decision mapping is (`yolt_analyzer.py:1200-1208`):

```
safe    -> exit silently, same as unknown
unsafe  -> permissionDecision: ask (deny inside a subagent)
unknown -> exit silently
```

`safe` and `unknown` are **indistinguishable to Claude Code**. So deleting the
per-CLI subcommand tables -- the bulk of what Phase 3 removes -- moves
commands from `safe` to `unknown`, which is a no-op at the permission prompt.
It changes the decision log and nothing else.

The one behavioral risk runs the other way: a command that is `unsafe` today
losing its classification and becoming `unknown`. That is an `ask` silently
disappearing, it is the only way this phase can change what the operator
experiences, and it is fully enumerable offline against a corpus that already
exists. Hence a replay, not a window.

## The agreement check earned its place on the first run

The harness prints, first, how often replaying the corpus reproduces the
decision the log recorded. The first run said **75%**, which reads as a broken
harness.

It was not. `_log_hook_decision` stores `command[:500]`, and **2,459 of 6,305
unique commands sit at exactly that cap**. A truncated compound command loses
its trailing segments, and those are disproportionately the mutating ones
(`&& git push`, `&& gh pr merge`), so the replay classifies a fragment:

| subset | reproduces its logged verdict |
| --- | --- |
| intact (< 500 chars) | **99.6%** (3830/3846) |
| truncated (== 500) | 54.9% (1339/2438) |

**1,099 of the 1,115 disagreements are that artifact alone.** Truncated
records are excluded by default and the count dropped is printed every run,
rather than being quietly averaged into a number that looks like a
classifier regression.

Filed separately as #119: the cap costs the gate 39% of its corpus, and
raising it is not free because redaction runs *before* truncation.

## Baseline

```
corpus:    3846 unique commands
           2459 truncated at 500 chars (excluding them)
agreement with logged decision: 3828/3846 (99.5%)
unsafe: 698 commands across 143 distinct reasons
```

That grouping by reason is #100 section 2's "capture the accumulated judgement
before deleting it", discharged as an artifact rather than an intention. It
also makes writing the non-delegable list a measured exercise: draft the list,
run `--diff baseline.json`, read exactly which asks you just gave up.

## Two deliberate details

**`--hooks-dir` is required and takes an arbitrary checkout.** The corpus was
produced by the umbrella build, not master, so a harness that imported its own
repo's classifier would silently measure the wrong thing. It also probes
`GrammarClassifier.__init__` for `allow_patterns`, because Phase 1 removed
that parameter and the whole point is comparing across that boundary.

**Allow-pattern resolution is off by default.** `run_hook` reads the
operator's real `~/.claude/settings.json`; resolving it here would make
results machine-specific, which is the hermeticity problem in #75.

## Not committed

The baseline JSON. It contains verbatim command lines including local paths,
and this repo is public. It lives at `~/.claude/yolt-phase3-baseline-<date>.json`.

## Version

None. `scripts/` is in the workflow's `paths-ignore`.

### DIFF
--- a/scripts/replay_unsafe.py
+++ b/scripts/replay_unsafe.py
@@ -1,0 +1,289 @@
+#!/usr/bin/env python3
+"""Replay a YOLT decision log through a classifier and report the `unsafe` set.
+
+This is the gate for Phase 3 (#100), and the artifact for its "capture the
+retired judgement before deleting it" requirement.
+
+WHY A REPLAY AND NOT A WINDOW
+
+  Post-Phase-1 the decision mapping is:
+
+      safe    -> exit silently, same as unknown
+      unsafe  -> permissionDecision: ask (deny inside a subagent)
+      unknown -> exit silently
+
+  `safe` and `unknown` are indistinguishable to the host. So deleting the
+  per-CLI subcommand tables -- the bulk of what Phase 3 removes -- moves
+  commands from `safe` to `unknown`, which is a no-op at the host. It changes
+  the log and nothing else.
+
+  The one behavioral risk is the opposite direction: a command that is
+  `unsafe` today losing its classification and becoming `unknown`. That is an
+  `ask` silently disappearing, and it is the only way Phase 3 can change what
+  the operator experiences. It is fully enumerable offline against a corpus
+  that already exists, so this needs no measurement window at all.
+
+USAGE
+
+  Baseline, against the build that produced the corpus:
+
+    python3 replay_unsafe.py \
+        --hooks-dir ~/g/git.voitta/voitta-yolt.worktrees/auto-mode-realignment/hooks \
+        --json > baseline.json
+
+  Candidate, against a Phase 3 branch, reporting what would lose its `ask`:
+
+    python3 replay_unsafe.py --hooks-dir <phase3>/hooks --diff baseline.json
+
+THE AGREEMENT CHECK IS LOAD-BEARING
+
+  The log records the decision each command actually received. Replaying
+  should reproduce it. The agreement rate is printed first and is the signal
+  that the harness is pointed at the right code: a baseline run that does not
+  agree with its own corpus is measuring some other build, and every number
+  after it is meaningless.
+
+  Allow-pattern resolution is OFF by default. `run_hook` reads the operator's
+  real `~/.claude/settings.json`, which would make results machine-specific
+  (the hermeticity problem in #75). Pass --allow-patterns to opt in.
+"""
+
+import argparse
+import collections
+import inspect
+import json
+import sys
+from pathlib import Path
+
+
+def load_corpus(log_path, cap, include_truncated):
+    """Unique commands from a YOLT decision log, with their logged decision.
+
+    Later records win: a command reclassified after a rules change should be
+    compared against the newer verdict, not the stale one.
+
+    Commands at exactly `cap` characters were TRUNCATED by the log writer
+    (`_log_hook_decision` stores `command[:500]`). They are excluded by
+    default and this is not a nicety: a cut compound command loses its
+    trailing segments, and those are disproportionately the mutating ones
+    (`&& git push`, `&& gh pr merge`). Replaying them reproduces the logged
+    verdict only 55% of the time against 99.6% for intact commands, so
+    including them buries a correct result under an artifact of the log
+    format.
+    """
+    seen = {}
+    truncated = 0
+    with open(log_path, errors="replace") as handle:
+        for line in handle:
+            line = line.strip()
+            if not line:
+                continue
+            try:
+                record = json.loads(line)
+            except ValueError:
+                continue
+            command = record.get("command")
+            decision = record.get("decision")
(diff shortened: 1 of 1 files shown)


scripts/replay_unsafe.py
```diff
@@ -0,0 +1,289 @@
...
+        print()
+        print(f"LOST unsafe vs {args.diff}: {len(lost)} commands")
+        print("Each one is an `ask` that silently disappears. Review every "
+              "shape before shipping.")
+        for command in lost[:40]:
+            flat = " ".join(command.split())
+            print(f"  {flat[:110]}")
+        if len(lost) > 40:
+            print(f"  ... {len(lost) - 40} more (use --json for the full set)")
+    return 0
```
debedb: **[critical] CI gate always exits 0 even on low agreement or lost unsafe** _(confidence 0.86)_

Load‑bearing safety checks are not enforced. Despite the docstring stating the agreement check is “load‑bearing,” the code only prints a warning when agreement < 99% and then continues. Likewise, even when a baseline diff detects lost unsafe decisions (silent removal of required asks), the program still exits 0. In CI this will green‑light Phase 3 even when the classifier is mis‑pointed (low agreement) or when previously‑unsafe commands have become non‑unsafe, creating real user‑visible risk that destructive operations proceed without an ask. This is a trust‑boundary failure of the gate itself, not the classifier: the script never fails closed.

**Recommendation:** Enforce gating with non‑zero exits. Concretely: 1) Add `--min-agreement` (default 99.5). After computing `pct`, if `total==0` or `pct < args.min_agreement`, print an error and `return 2`. 2) When `--diff` is provided, if `lost` is non‑empty, print a concise summary and `return 3` (also include `lost_unsafe_count` in JSON mode). 3) Optionally add `--no-fail-on-lost` for exploratory runs, but default to failing. 4) Propagate the same failure codes in JSON mode instead of always `return 0`. This makes the script a real CI gate and prevents shipping when the only load‑bearing precondition (high agreement) or the core safety signal (lost unsafe) fails.

debedb (commented):

Do not ship. The script that claims to be the Phase 3 gate never fails closed: it prints a warning on low agreement and even when unsafe decisions are lost, then exits 0. This will pass CI while silently removing required asks.

### Next steps
- Add `--min-agreement` (default 99.5) and fail with exit 2 when breached.
- Fail with exit 3 when `--diff` is set and any `lost_unsafe` exist; expose `lost_unsafe_count` in JSON.
- Wire these exits in both text and `--json` paths; update usage docs accordingly.

debedb:
Reviewed the finding against the tree. **Half of it is right and is now
fixed; the other half rests on a premise that does not hold.**

## Taken

**Low agreement now fails closed.** `--min-agreement` (default 99.0) exits 2.
The reviewer is correct that a printed warning followed by `return 0` is not a
gate — a run below the threshold reproduced its own corpus badly, which means
the classifier is not the build that produced the log, and every number after
that describes something else. Returning 0 invited a caller to consume noise.

Verified in both directions rather than asserted:

```
against the umbrella build that produced the corpus -> 99.5%, exit 0
deliberately mis-pointed at master (pre-Phase-1)    -> exit 2
```

Also taken: `lost_unsafe_count` in the JSON payload, and the same exit codes
on the JSON path instead of an unconditional `return 0`.

## Declined, with reasons

**"This will pass CI while silently removing required asks" is false.** The
script is not in CI and cannot green-light anything. `scripts/**` appears in
`.github/workflows/release.yml` exactly once, as a `paths-ignore` entry — the
version-bump exemption. No workflow executes anything under `scripts/`;
`ci.yml` runs `python -m unittest discover -v tests` and nothing else. The
`critical` severity and the "Do not ship" verdict both rest on wiring that
does not exist.

**Exiting non-zero whenever `--diff` finds a lost unsafe would break the
tool.** Phase 3 exists *in order to* delegate commands to the host, so a
non-empty `lost_unsafe` is the expected output of every legitimate `--diff`
run, not an error condition. Making it exit 3 by default means the tool
reports failure on success. The tell is in the recommendation itself: it
proposes `--no-fail-on-lost` to escape the failure mode it introduces.

The review step this feeds is a human judgement — #100: "Phase 3 ships when
each shape in that list has been read and accepted as delegable" — and an exit
code cannot stand in for reading the list. `--fail-on-lost` is now available
opt-in for a caller that genuinely wants zero tolerance.

## Note on the review process

This is the "remedy disproportionate to the defect" shape the
`codex-adversarial-pr-review` skill documents: a confident, well-written
`critical` at 0.86 whose central factual claim is refuted by one `grep` of
`.github/workflows/`. Posting it unedited was deliberate — the valid half is
genuinely valuable and the record should show what the reviewer actually said
— but it is worth logging as another instance of the pattern.

PR #122, opened 2026-08-29
Phase 4: reuse #82's git predicates deny-side, with the failure semantics inverted

### LINKED ISSUE #97: Reuse #82's git-state predicates deny-side: first entries in the non-delegable list

debedb:
Closes #97. Targets the umbrella, not master. No version bump.

## What carried over, and what could not

#82 built three deterministic git-state predicates and wired them to **grant**
approval. The predicates are good; the direction was wrong. This keeps them
and inverts them.

Ported as-is: `GitProbe`, the literal-`cd` tracker, `git -C` retargeting,
per-directory memoization, schema validation.

**The failure handling could not be ported, and that is the substance of this
PR.** #97 flagged this as the one place a verbatim port would be actively
wrong. It was more concrete than that:

`solo_author` returned `(False, ...)` for *both* "someone else authored these
commits" and every probe failure — `user.email` unset, `origin/master` never
fetched, unreadable history. Granting, that conflation is safe: any False
means no grant and the `ask` stands. **Denying, a repo with no configured
`user.email` would deny every force push in it**, and a git outage would deny
everything.

So authorship is tri-state here — `SOLO` / `SHARED` / `UNDETERMINED` — and
every probe failure resolves to `UNDETERMINED`, which produces no opinion.

The invariant, asserted by nine tests: **no probe failure of any kind can
produce a deny.**

## Entries chosen against the corpus, not guessed

Measured with `scripts/replay_unsafe.py` over the 12-day dogfood corpus:

**1. Push at the default branch** — enabled. Fired **zero times across 76
observed pushes**; every one targeted a feature branch. A net safety guard
with no measured friction.

It still needs a probe, and this is the case for the whole issue: **24 of
those 76 pushes carried no refspec at all** — including `git push -f` and
`git push -qf`. They push whatever is checked out, and argv cannot know what
that is. Only `rev-parse --abbrev-ref HEAD` can.

**2. Force-push or delete over commits somebody else authored** — enabled.
`--force-with-lease` is deliberately in scope: it guards against a moved
remote, not against overwriting a coauthor whose commits you already have
locally. It was 32 of the 76 pushes, so this is the shape that matters.

**3. Commits from the primary checkout** — **disabled**. A house convention,
not a universal truth, so it ships off.

## One deviation from the issue text, flagged deliberately

#97 specifies that a probe failure "must fall through to `unknown`". This
implements *no deny; the pre-existing decision stands*, which for these
commands means `unsafe` → `ask`.

Reasoning: falling to `unknown` degrades toward **less** friction, since
`unknown` exits silently post-Phase-1. Leaving `unsafe` in place degrades
toward **more**, which is the direction #82's failure handling went and the
direction the issue's own argument favours. The invariant the issue actually
cares about — a failed probe never denies — holds either way. If you want the
literal wording, it is a one-line change in `_maybe_deny_by_policy`.

## Decision vocabulary

`DECISION_DENY` is a fourth decision at the top of precedence (deny > unsafe >
unknown > safe), reachable **only** by upgrading something already classified
`unsafe`. Nothing reaches deny from `safe` or `unknown`. Deny outranks its
siblings because one refused segment condemns the line — there is no way to
run `A && B` with B refused.

## Verified against live git, not only fixtures

| command | from | result |
| --- | --- | --- |
| `git push` | master checkout | **DENY** (default branch) |
| `git push -f` | master checkout | **DENY** |
| `git push` | feature worktree | ask |
| `git push --force-with-lease` | feature worktree, solo | ask |
| `git push origin master` | feature worktree | **DENY** |
| `cd <master> && git push` | feature worktree | **DENY** (cd tracked) |
| `git -C <master> push` | feature worktree | **DENY** (-C retargeted) |
| `git push origin $BRANCH` | master checkout | no deny (unresolvable) |
| `git push` | `/tmp`, not a repo | ask (probe failed) |
| `git push` + `YOLT_NO_POLICIES=1` | master checkout | ask |

687 tests pass, up from 659 — 28 new, none broken.

## Worth knowing for review

Three wiring bugs got through the unit tests, all the same shape: I patched
against the pre-Phase-1 code from #82's diff, and Phase 1 had deleted the
anchors. `DECISION_DENY` went unimported in two modules and the `cd` hook
never landed — and every one of those passed the policy unit tests, because
those call `GitDenyPolicy` directly and never touch the wiring. The
`ThroughTheClassifier` test class exists specifically to close that gap and
was added after the end-to-end run caught what the units could not.

### DIFF
--- a/hooks/git_policy.py
+++ b/hooks/git_policy.py
@@ -334,6 +334,10 @@ class GitDenyPolicy:
             targets = _push_targets(argv, state.get("branch"))
             if not targets:
                 return None
+            if "*" in targets:
+                return ("pushes every branch (--all/--mirror), which "
+                        "includes the default branch ({})".format(
+                            default_branch))
             if default_branch in targets:
                 return "would push to the default branch ({})".format(
                     default_branch
@@ -371,21 +375,27 @@ def _push_targets(argv, current_branch):
     """Every branch a `git push` would land on, or None if argv cannot say.
 
     Returns a set. A push can name several refspecs at once, and taking only
-    the last one -- as an earlier version of this did -- lets
-    `git push origin master feature/x` slip past the default-branch guard,
-    because the guard only ever looked at `feature/x`.
+    the last one -- as the first version of this did -- let
+    `git push origin master feature/x` slip past the default-branch guard.
 
-    With no refspec, git pushes the current branch, so the answer comes from
-    repository state rather than the command. None means unresolvable, which
-    denies nothing:
+    The bulk flags are not interchangeable and must not be handled as one
+    group, which was the second bug here:
 
-    - `--all` / `--mirror` / `--tags` push a whole namespace rather than a
-      named branch, so the destination set is not derivable from argv.
-    - a variable, a glob, or a `refs/` ref outside `refs/heads/`.
+    - `--all` and `--mirror` push every branch, so they necessarily include
+      the default branch when it exists. They are a target, not an unknown.
+    - `--tags` pushes tags *in addition to* whatever refspecs are given. It
+      says nothing about branches, so it must not suppress parsing of an
+      explicit `master` sitting next to it -- `git push origin master --tags`
+      is still a push at the default branch.
+
+    With no refspec and no bulk flag, git pushes the current branch, so the
+    answer comes from repository state rather than the command. None means
+    unresolvable, which denies nothing.
     """
-    for bulk in ("--all", "--mirror", "--tags"):
-        if bulk in argv:
-            return None
+    if "--all" in argv or "--mirror" in argv:
+        # Pushes every branch. The default branch is in that set by
+        # definition, so this resolves rather than being unknown.
+        return {current_branch, "*"} if current_branch else {"*"}
 
     positionals = [
         tok for tok in argv[1:]
@@ -393,6 +403,9 @@ def _push_targets(argv, current_branch):
     ]
     refspecs = positionals[1:]
     if not refspecs:
+        if "--tags" in argv:
+            # Tags only, no branch touched.
+            return None
         return {current_branch} if current_branch else None
 
     targets = set()
@@ -473,23 +486,3 @@ def _resolve_directory(c_target, directory):
 
 def _has_expansion(token):
     return "$" in token or "`" in token or "*" in token
-
-
-def _pushes_elsewhere(argv, branch):
-    """True when a `git push` names a ref other than the current branch.
-
-    `git push` and `git push origin <current-branch>` stay inside the
-    policy; `git push origin master` or any `src:dst` refspec is a push at
-    something the predicates never examined, so it falls back to `ask`.
-    """
-    positionals = [
-        tok for tok in argv[1:]
-        if not tok.startswith("-") and tok != "push"
-    ]
-    # positionals[0] is the remote; anything after it is a refspec.
-    for refspec in positionals[1:]:
-        if ":" in refspec:
-            return True
-        if refspec != branch:
-            return True
-    return False
--- a/hooks/git_policy.py
+++ b/hooks/git_policy.py
@@ -1,0 +1,479 @@
+"""Policy layer: upgrade `unsafe` to `deny` for git writes we refuse outright.
+
+The rest of YOLT is a static checker -- it answers "does this command mutate
+anything?" from syntax alone. That question has no good answer for `git push`:
+pushing a feature branch you opened five minutes ago and force-pushing over
+someone else's commits on the default branch are different acts, and the argv
+can be byte-identical. Measured on a 12-day corpus, 24 of 76 observed pushes
+carried no refspec at all -- including `git push -f` -- so they push whatever
+happens to be checked out, which argv cannot know.
+
+So this module answers a different question -- "is this write one we refuse to
+delegate to a probabilistic classifier?" -- and it needs repository state to do
+it, which means shelling out to `git`. That is a deliberate, scoped exception
+to the no-subprocess design principle:
+
+- It runs only when the static classifier already said `unsafe` AND the argv
+  head matches a configured policy entry, so the common path never forks.
+- It can only turn `unsafe` into `deny`. It never makes a safe command unsafe,
+  and it never sees a command the classifier did not already flag.
+- Every probe is read-only `git`, with a timeout.
+
+INVERTED FAILURE SEMANTICS -- the one thing that could not be ported
+
+Its ancestor (#82) used these same probes to *grant* approval, where every
+probe failure meant "predicate not satisfied", the grant did not happen, and
+the original `ask` stood. Failure degraded toward more friction, which is
+safe, so that code was free to collapse "definitely not" and "cannot tell"
+into a single False.
+
+Denying inverts that. Collapsed the same way, an unset `git config user.email`
+or an unfetched `origin/master` reads as "not solo" and would **deny every
+force push in the repository**. A git outage or a slow network filesystem
+would brick the session. That is the failure mode #97 names, and porting the
+predicate verbatim would have shipped it.
+
+So authorship is tri-state here -- SOLO / SHARED / UNDETERMINED -- and every
+probe failure resolves to UNDETERMINED, which produces no opinion. The
+invariant, asserted by the tests: **no probe failure can ever produce a deny.**
+"""
+
+import os
+import subprocess
+
+# Read-only probes, but a hung git (network filesystem, index.lock
+# contention, a filter process) must not hang the permission prompt.
+_GIT_TIMEOUT_SECONDS = 3
+
+_DEFAULT_BRANCH_FALLBACKS = ("main", "master")
+
+# Authorship states. UNDETERMINED is not a failure to be retried or a
+# pessimistic default -- it is a first-class answer meaning "no opinion", and
+# it is what every probe failure resolves to.
+SOLO = "solo"
+SHARED = "shared"
+UNDETERMINED = "undetermined"
+
+# Predicate names a policy entry may list in its `refuse_when` array. Each is
+# a condition under which the command is DENIED. A predicate that cannot be
+# evaluated never fires.
+PREDICATES = ("default_branch_target", "shared_history", "primary_checkout")
+
+
+class GitProbe:
+    """Read-only `git` queries about one directory, memoized per instance.
+
+    One hook invocation classifies one Bash command, but that command may
+    be `git add ... && git commit ... && git push`, which asks the same
+    questions of the same directory three times. Memoizing keeps that at
+    one round of probes.
+    """
+
+    def __init__(self, runner=None):
+        # Injectable for tests; signature mirrors `_run_git`.
+        self._runner = runner or _run_git
+        self._cache = {}
+
+    def state(self, directory):
+        """Return a dict of facts about `directory`, or None if it is not
+        usable as a git working tree."""
+        if directory is None:
+            return None
+        key = os.path.abspath(directory)
+        if key not in self._cache:
+            self._cache[key] = self._probe(key)
+        return self._cache[key]
+
+    def _probe(self, directory):
+        if not os.path.isdir(directory):
+            return None
+
+        out = self._runner(
+            directory,
+            ["rev-parse", "--git-dir", "--git-common-dir", "--abbrev-ref", "HEAD"],
+        )
+        if out is None:
+            return None
+        lines = out.splitlines()
+        if len(lines) < 3:
+            return None
+        git_dir, git_common_dir, branch = lines[0], lines[1], lines[2]
+
+        # In a linked worktree these differ: --git-dir points at
+        # <main>/.git/worktrees/<name> while --git-common-dir points at
+        # <main>/.git. In the primary checkout they are the same path.
+        linked_worktree = os.path.abspath(
+            os.path.join(directory, git_dir)
+        ) != os.path.abspath(os.path.join(directory, git_common_dir))
+
+        return {
+            "directory": directory,
+            "branch": branch,
+            "detached": branch == "HEAD",
+            "linked_worktree": linked_worktree,
+            "default_branch": self._default_branch(directory),
+            "user_email": self._config(directory, "user.email"),
+        }
+
+    def _default_branch(self, directory):
+        # The remote's HEAD is the authoritative answer when the remote
+        # has been fetched at least once.
+        out = self._runner(
+            directory, ["symbolic-ref", "--short", "refs/remotes/origin/HEAD"]
+        )
+        if out:
+            head = out.strip()
+            if head.startswith("origin/"):
+                return head[len("origin/"):]
+
+        # No remote HEAD (never fetched, or no remote). Fall back to
+        # whichever candidate actually resolves in THIS repo.
+        # `init.defaultBranch` is only a candidate, never an answer on its
+        # own: it is a global preference for repos yet to be created, so a
+        # user who sets it to `main` would otherwise make every existing
+        # `master` repo look like it has no default branch.
+        candidates = []
+        configured = self._config(directory, "init.defaultBranch")
+        if configured:
+            candidates.append(configured)
+        candidates.extend(
+            c for c in _DEFAULT_BRANCH_FALLBACKS if c not in candidates
+        )
+
+        for candidate in candidates:
+            for ref in ("refs/remotes/origin/" + candidate, "refs/heads/" + candidate):
+                if self._runner(
+                    directory, ["rev-parse", "--verify", "--quiet", ref]
+                ) is not None:
+                    return candidate
+        return None
+
+    def _config(self, directory, key):
+        out = self._runner(directory, ["config", "--get", key])
+        if not out:
+            return None
+        return out.strip() or None
+
+    def authorship(self, directory):
+        """SOLO / SHARED / UNDETERMINED for the commits on this branch.
+
+        Returns a (state, detail) pair. SOLO means every commit on this
+        branch that is not on the default branch was authored AND committed
+        by the configured user. SHARED means at least one was not.
+        UNDETERMINED means the question could not be answered, and is
+        returned for every probe failure -- no git, no user.email, no
+        comparable ref, unreadable history.
+
+        The distinction between SHARED and UNDETERMINED does not exist in
+        the ancestor of this module, and adding it is the whole point of the
+        port: see the failure-semantics note in the module docstring.
+
+        A branch with no commits of its own yet is vacuously solo — that
+        is the state you are in the moment before the first `git commit`,
+        which is exactly the case this policy exists to allow.
+        """
+        state = self.state(directory)
+        if state is None:
+            return (UNDETERMINED, "not a git working tree")
+
+        email = state.get("user_email")
+        if not email:
+            return (UNDETERMINED, "git user.email is not configured")
+
+        default_branch = state.get("default_branch")
+        if not default_branch:
+            return (UNDETERMINED, "cannot determine the default branch")
+
+        base = None
+        for ref in ("origin/{}".format(default_branch), default_branch):
+            if self._runner(
+                directory, ["rev-parse", "--verify", "--quiet", ref]
+            ) is not None:
+                base = ref
+                break
+        if base is None:
+            return (UNDETERMINED,
+                    "no {} ref to compare against".format(default_branch))
+
+        out = self._runner(
+            directory, ["log", "--format=%ae%n%ce", "{}..HEAD".format(base)]
+        )
+        if out is None:
+            return (UNDETERMINED, "cannot read branch history")
+
+        others = {
+            line.strip() for line in out.splitlines()
+            if line.strip() and line.strip() != email
+        }
+        if others:
+            return (SHARED, "branch has commits by {}".format(
+                ", ".join(sorted(others))
+            ))
+        return (SOLO, "every commit on this branch is {}".format(email))
+
+
+class DenyPolicySet:
+    """Every configured deny policy, sharing one probe cache."""
+
+    def __init__(self, policies):
+        probe = GitProbe()
+        self.policies = [
+            GitDenyPolicy(spec, probe=probe)
+            for name, spec in sorted((policies or {}).items())
+            if not name.startswith("_") and isinstance(spec, dict)
+        ]
+
+    def evaluate(self, argv, directory):
+        for policy in self.policies:
+            reason = policy.evaluate(argv, directory)
+            if reason is not None:
+                return reason
+        return None
+
+
+def load_policies(rules):
+    """Build the deny policy set from loaded shell rules. Returns None when
+    nothing is configured, so the classifier can skip the layer entirely."""
+    policy_set = DenyPolicySet(rules.get("policies", {}))
+    if not any(p.enabled and p.entries for p in policy_set.policies):
+        return None
+    return policy_set
+
+
+def _run_git(directory, args):
+    """Run a read-only git command. Returns stdout, or None on any failure.
+
+    A failed probe must never read as evidence. On the deny side that means
+    None propagates to UNDETERMINED and the command is not denied.
+    """
+    try:
+        proc = subprocess.run(
+            ["git", "-C", directory, "--no-pager"] + args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            timeout=_GIT_TIMEOUT_SECONDS,
+        )
+    except (OSError, subprocess.SubprocessError):
+        return None
+    if proc.returncode != 0:
+        return None
+    return proc.stdout.decode("utf-8", "replace")
+
+
+class GitDenyPolicy:
+    """Evaluates configured deny entries against an argv + directory."""
+
+    def __init__(self, spec, probe=None):
+        spec = spec or {}
+        self.enabled = bool(spec.get("enabled", False))
+        if os.environ.get("YOLT_NO_POLICIES"):
+            self.enabled = False
+        self.entries = [e for e in spec.get("deny", []) if isinstance(e, dict)]
+        self.probe = probe or GitProbe()
+
+    def evaluate(self, argv, directory):
+        """Return a reason string when this argv must be DENIED, else None.
+
+        None is "no opinion", and it is the answer for every uncertainty:
+        policy disabled, argv unmatched, directory unresolvable, probe
+        failed, predicate undetermined. Callers only invoke this for argv
+        the static rules already classified `unsafe`, so a None here leaves
+        that `unsafe` standing -- the operator still gets asked.
+        """
+        if not self.enabled or not argv:
+            return None
+
+        # `git -C <path>` retargets the command; the policy must judge the
+        # directory git will actually act on, not the shell's cwd.
+        argv, c_target = _split_dash_c(argv)
+
+        # Every matching entry is evaluated, not just the most specific
+        # one. Two entries legitimately share the `git push` prefix -- one
+        # refusing pushes at the default branch, one refusing force-pushes
+        # over shared history -- and picking a single winner would silently
+        # disable whichever lost.
+        entries = self._match_all(argv)
+        if not entries:
+            return None
+
+        directory = _resolve_directory(c_target, directory)
+        if directory is None:
+            return None
+
+        state = self.probe.state(directory)
+        if state is None:
+            return None
+        # Detached HEAD has no branch to reason about. Not an error, but
+        # nothing here can be established, so no opinion.
+        if state.get("detached"):
+            return None
+
+        for entry in entries:
+            # An entry may scope itself to particular flags, so that
+            # `git push --force` is refused while a plain push is not.
+            only_flags = entry.get("only_flags")
+            if only_flags and not any(flag in argv for flag in only_flags):
+                continue
+            label = " ".join(entry.get("argv", []))
+            for predicate in entry.get("refuse_when", []):
+                reason = self._fires(predicate, argv, state, directory)
+                if reason is not None:
+                    return "{}: {}".format(label, reason)
+        return None
+
+    def _fires(self, predicate, argv, state, directory):
+        """Return a reason when `predicate` is affirmatively true, else None.
+
+        Every branch that cannot establish the fact returns None rather than
+        guessing. That is the fail-open invariant in one place.
+        """
+        if predicate == "default_branch_target":
+            default_branch = state.get("default_branch")
+            if not default_branch:
+                return None
+            target = _push_target(argv, state.get("branch"))
+            if target is None:
+                return None
+            if target == default_branch:
+                return "would push to the default branch ({})".format(
+                    default_branch
+                )
+            return None
+
+        if predicate == "shared_history":
+            verdict, detail = self.probe.authorship(directory)
+            if verdict == SHARED:
+                return detail
+            return None
+
+        if predicate == "primary_checkout":
+            if state.get("linked_worktree") is False:
+                return "runs in the primary checkout, not a linked worktree"
+            return None
+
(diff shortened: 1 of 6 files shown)


debedb (commented):

Do not ship. The deny-side git policy depends on scattered None checks and a narrow refspec heuristic; a missed guard or real-world refspec form can escalate to an unintended DENY, bricking pushes. Repo resolution and global DENY precedence can misattribute policy decisions across substitutions and directories. Harden parsing, centralize failure semantics, and extend tests before release.

### Next steps
- Add fuzz/timeout tests for GitProbe runner returning partial/garbled outputs and assert DENY never occurs.
- Replace _push_target with a refspec-normalizing parser and deny-if-any-destination==default semantics; add tag/notes/wildcard/HEAD^/@{u} tests.
- Tighten repo resolution: verify resolved dir is a git worktree toplevel; treat ambiguity as UNDETERMINED with an explicit reason.
- Refactor aggregation to segment-scoped precedence and add tests covering command substitutions and mixed pipelines.
- Enable minimal debug observability on probe failures (opt-in env flag) to aid postmortems without leaking data.

- `hooks/git_policy.py:1-120` **[high] Deny-on-failure invariant relies on scattered None checks; a missed guard can spuriously DENY under probe timeouts/errors** _(confidence 0.43)_
  Timeouts, OSErrors, or non‑zero exits in read‑only git probes can be misinterpreted as definitive authorship checks if any return path fails to propagate “undetermined.” Specifically: (1) _default_branch() returns None when fetch state is missing or git is slow; (2) authorship() then short‑circuits to UNDETERMINED; but (3) default_branch_target uses _push_target() which falls back to current branch, and if state['branch'] itself came from a partially parsed rev‑parse, an empty or truncated branch string can compare equal to default_branch under odd remap conditions; (4) shared_history predicate relies on a log scan that can silently truncate under timeout and could produce a non‑empty set due to partial decoding. While tests assert “no probe failure can ever produce a deny,” the code depends on None checks scattered across multiple helpers rather than a single, explicit failure sentinel. Under load (network FS, filter drivers), any missed None guard would elevate DECISION_DENY unexpectedly, bricking pushes. A single guard bug in any helper violates the invariant and will be expensive and hard to diagnose because stderr is suppressed (stderr=DEVNULL) and reasons collapse to policy labels. _(fix: Introduce an explicit ProbeResult object with status {OK, UNDETERMINED} carrying all derived fields; make evaluate() return early on any UNDETERMINED without consulting predicates. Centralize timeout/exception handling and log minimal diagnostics (e.g., reason codes) surfaced alongside policy reasons. Add tests that fuzz runner to return empty strings, partial lines, and encoding errors to ensure no DENY can arise. Remove stderr=DEVNULL in probes or gate it behind a debug flag to keep observability.)_
- `hooks/grammar_classifier.py:200-330` **[medium] cd/-C tracking can target the wrong repo under real-world shells** _(confidence 0.50)_
  Directory tracking can judge the wrong repository. The policy evaluates git against `_cwd`, updated only by literal `cd` and `git -C`. However: (1) `cd` without args is treated as HOME rather than POSIX default `$HOME` but doesn’t verify the path exists; (2) relative `-C` targets resolve even if they point to a non-git directory (state() later returns None, but before that the refspec match may allow predicates to run based on the caller’s argv assumptions); (3) nested worktrees or submodules aren’t distinguished from unrelated parent repos if the shell starts in a subdir that isn’t actually the intended worktree; (4) expansions in `cd` beyond simple $, `*`, `` ` `` are not considered (e.g., brace expansion), leading to stale `_cwd` that still triggers policy in the previous repo. These edge cases create false DENY or missed DENY depending on where the accidental repository points. _(fix: Harden directory resolution: (a) require existence checks for HOME fallback; (b) validate that the resolved directory is inside a git working tree before evaluating any predicate; (c) treat submodule and worktree boundaries explicitly by calling `git rev-parse --show-toplevel` and comparing to intended dir; (d) expand brace patterns conservatively by treating any non-literal `cd` as UNDETERMINED; (e) add tests for `cd` with no args, non-existent paths, submodules, nested worktrees, and chained `-C` plus `cd` transitions.)_
- `hooks/rule_classifier.py:92-118` **[medium] Global DENY precedence may overreach across substitutions and segments** _(confidence 0.46)_
  DECISION_DENY precedence across pipelines and substitutions risks unintended elevation when a denied segment coexists with opaque or unknown segments. aggregate_decisions gives deny > unsafe > unknown > safe globally, and GrammarClassifier separately walks command_substitution/process_substitution. If a substitution executes an unrelated write classified as unsafe/unknown while the primary command matches a deny policy, the whole command is DENY — even when the substitution is logically independent of the git push the policy targets. This makes policy scope bleed into unrelated nested commands (e.g., `cmd=$(touch /tmp/x; echo a); git push`). While tests cover one precedence case, they don’t cover command substitution interactions or multiple primary commands split by `;` or `&&` with mixed substitutions, increasing the chance of accidental DENY escalation beyond the intended git write refusal. _(fix: Constrain policy-applied DENY to the specific command segment that triggered it instead of global collapse: aggregate per-primary-command first, then combine at the list level so a DENY from policy only condemns the primary command it judged. Add tests for command substitution containing writes coexisting with a denied git push; verify that the DENY reason remains attributable and that unrelated segments don’t spuriously flip from ASK to DENY.)_

</details>

hooks/git_policy.py
```diff
@@ -0,0 +1,479 @@
...
+            continue
+        if tok != remaining[0]:
+            return False
+        remaining.pop(0)
+    return not remaining
+
+
+def _split_dash_c(argv):
+    """Return (argv without any `-C <path>` pair, the last such path).
+
```
debedb, line 449: **[high] Refspec parsing under-specifies real git forms; wrong target can DENY or miss DENY** _(confidence 0.62)_

_push_target parsing is too permissive about refspec forms and collapses multiple refspecs by returning the last one. Real git allows multiple refspecs, wildcard refspecs, and tag or notes namespaces; a push containing SRC:DST plus additional refspecs or a push involving tags (e.g., --tags) will be misread. The code strips refs/heads/ but treats any other refs/* (including refs/tags/*) as unresolvable -> None, silently skipping an explicit default-branch target if it’s spelled with a different namespace or an abbreviated form. It also treats SUBSTITUTION_PLACEHOLDER or values with at-signs and caret syntax (e.g., HEAD^, branch@{u}) as plain strings, leading to false positives/negatives. Result: default_branch_target may fail open or mis‑target, and shared_history may be evaluated for the wrong destination.

**Recommendation:** Implement refspec parsing aligned with git’s grammar: handle multiple refspecs deterministically (deny if any destination equals default branch), support heads/tags/notes namespaces explicitly, and reject pushes with --all/--mirror/--tags or wildcard refspecs from policy evaluation (treat as UNDETERMINED). Normalize destinations via `git rev-parse --symbolic-full-name` when possible within the UNDETERMINED guard. Add tests covering: multiple refspecs, wildcard, tags, notes, HEAD^, @{u}, and SUBSTITUTION_PLACEHOLDER cases.

debedb:
Checked each finding against the tree. **One held up and is fixed; three do
not hold.** Detail below, since the three all turn on the same fail-open
property and it is worth showing rather than asserting.

## Fixed: every refspec counts, not just the last

Correct, and a real miss-deny. `_push_target` returned the last refspec it
saw:

```
git push origin master feature/x   ->  'feature/x'
```

so the default-branch guard never saw `master`. It is now `_push_targets`,
returns a set, and the guard denies if **any** destination is the default
branch. Bulk flags (`--all`, `--mirror`, `--tags`) joined the unresolvable
set at the same time — they push a namespace, not a named branch, so argv
cannot yield the destinations. `refs/heads/<branch>` is normalised; other
`refs/` namespaces are skipped, since a tag ref has nothing to compare
against a branch name.

Tests added for multiple refspecs, `refs/heads/`, tag refs, and each bulk
flag. 691 pass.

## Does not hold: `HEAD^`, `@{u}`, substitution placeholders

These parse as themselves, compare unequal to the default branch, and
therefore deny nothing:

```
git push origin HEAD^        ->  'HEAD^'
git push origin branch@{u}   ->  'branch@{u}'
```

The finding frames them as "false positives/negatives". They can only ever
be negatives, which is the safe direction and the one this whole module is
built around.

## Does not hold: a truncated branch string could compare equal to the default

```
"" == "master"   ->  False
```

An empty or truncated branch cannot equal a default branch name, and
`_fires` already returns early when `default_branch` is falsy. There is no
input that reaches a deny this way.

## Does not hold: a log scan truncating under timeout could produce a false SHARED

`subprocess.TimeoutExpired` is a subclass of `subprocess.SubprocessError`,
which `_run_git` catches — returning `None`, which `authorship` resolves to
`UNDETERMINED`. A timeout cannot produce `SHARED`.

The broader recommendation attached to it — replace the None checks with a
`ProbeResult` sentinel object — is a reasonable design preference, but it is
offered as a fix for a failure mode that does not exist. The invariant is
asserted directly by nine tests that drive each failure path.

## Declined by design: segment-scoped DENY precedence

The finding argues that `cmd=$(touch /tmp/x); git push` denying the whole
line is overreach, and proposes per-segment precedence so a denied segment
does not condemn its siblings.

That inverts the guarantee. The unit of a `PreToolUse` decision is the whole
Bash invocation — there is no mechanism to run `A && B` with B refused. If
the push must be refused, refusing the line is the only correct answer;
per-segment precedence would let the denied push execute because a sibling
was fine. Deny outranking its siblings is deliberate and is stated in the
commit message.

## On the review itself

The first run returned a degenerate **plan-only** payload — a
`needs-attention` verdict whose sole finding was "I'm about to diff the
branch." I did not post it. The second run, with an explicit `--focus`
naming the four risk areas, produced the review above. Worth noting for the
`codex-adversarial-pr-review` skill: the plan-only shape it documents recurs,
and `--focus` appears to be the lever that gets past it.

One finding at 0.62 that turned out to be a real bug is a good result. The
three that did not hold were all in the collapsed sub-0.6 section, which is
the confidence floor working as intended.

hooks/git_policy.py
```diff
@@ -0,0 +1,495 @@
...
+        key = os.path.abspath(directory)
+        if key not in self._cache:
+            self._cache[key] = self._probe(key)
+        return self._cache[key]
+
+    def _probe(self, directory):
+        if not os.path.isdir(directory):
+            return None
+
+        out = self._runner(
```
debedb, line 91: **[high] Bulk-flag short‑circuit can suppress default‑branch deny for mixed pushes** _(confidence 0.62)_

Default-branch guard now denies if ANY target matches, but _push_targets treats any refs outside refs/heads as unresolvable and returns None. A mixed push that includes a default-branch ref and a tag or notes ref like `git push origin master refs/notes/commits` will yield targets={'master'} (ok). However, a push where the last positional is an unresolvable ref and earlier includes the default branch but is normalized away by an earlier bulk-like token is possible if flags are interleaved positionally: e.g. `git push origin master --tags`. The current parser gathers `positionals = [tok for tok in argv[1:] if not tok.startswith('-') and tok != 'push']` and then `refspecs = positionals[1:]`. `--tags` is excluded (starts with '-') so refspecs=['master']; but bulk flags earlier could still be present (e.g. `git push --mirror origin master`), and the top-of-function bulk guard only checks `if bulk in argv: return None` before positional extraction. Because that guard fires on presence anywhere, this case returns None and the guard never runs, silently allowing a simultaneous default-branch push bundled with a bulk push. This widens attack surface compared to “last refspec only” but still permits a deny to be skipped when it should fire (false negative) or, if future flags are extended, could produce inconsistent behavior across different orderings.

**Recommendation:** Tighten bulk-flag handling to consider refspec presence: if both a bulk flag and any concrete branch refspecs are present, compute targets from the concrete refspecs (do not return None). Specifically: (1) split argv into flags and positionals first; (2) if any of --all/--mirror/--tags are present AND there exists at least one explicit refspec in refs/heads or bare branch tokens, ignore the bulk flags for target computation, but still treat truly bulk-only cases as unresolvable. Add tests covering `git push --mirror origin master`, `git push --tags origin master feature/x`, and `git push origin master --tags`.

debedb (commented):

Do not ship. The new ANY-target deny is directionally right but has edge-path holes and ambiguity that can skip critical denies or create inconsistent scope. Mixed bulk flags suppress target derivation; empty-set conflation hides parsing regressions; refs/heads case-sensitivity can bypass normalization; primary-checkout relies on brittle state shape; and default-branch deny can fire on out-of-scope mapped pushes. These are fixable with small, testable guard refinements.

### Next steps
- Add targeted unit tests in tests/test_git_policy.py covering mixed bulk+explicit refspecs, None vs empty-set behavior, case-insensitive refs/heads, missing linked_worktree key, and mapped-push gating.
- Adjust _push_targets and _fires per recommendations; re-run the existing suite to ensure probe-failure invariant remains true.

- `hooks/git_policy.py:51-91` **[medium] Empty-set vs None conflation can mask target parsing errors** _(confidence 0.55)_
  _push_targets returns an empty set as None (`retval = targets or None`). In _fires, `targets = _push_targets(...); if not targets: return None`. This conflates two different states: (a) legitimately parsed but empty target set (e.g., all provided refspecs filtered out because they normalize to non-branch refs), and (b) truly unresolvable (variables, globs, bulk flags). In case (a), argv carried explicit refspecs that were parseable, and empty means “no heads among them” — arguably safe to not deny. But it also means a mixed list like `git push origin refs/tags/v1 refs/heads/master^{} ` could be parsed oddly by future ref forms and collapse to empty under normalization bugs, causing the guard to skip. More importantly, using falsy check means any future ref parsing that accidentally yields empty set will silently skip default-branch checks, hiding regressions. _(fix: Return a three-state result from _push_targets: (set, Unresolvable, ParsedEmpty). Or simpler: keep None for truly unresolvable, but return an explicit empty set when refspecs were present and all filtered to non-heads; in _fires, distinguish `targets is None` (skip) from `len(targets)==0` (no branch targets). Update tests to assert these branches. At minimum, replace `if not targets:` with `if targets is None:` in _fires.)_
- `hooks/git_policy.py:51-91` **[medium] Case-sensitive refs/heads normalization can be bypassed on case-sensitive filesystems** _(confidence 0.47)_
  Refs normalization skips non‑heads (`elif candidate.startswith('refs/'):` continue) but allows raw `master` etc. There is no case-normalization and no `heads/` typo handling. A malicious remote or unusual tooling could create a branch named `Refs/Heads/Master` (case sensitive FS) or a namespace collision where an explicit `refs/Remotes/origin/master` is passed in unusual casing via a script. While git’s ref namespaces are case-sensitive on some systems and normalized by git, your parser’s plain prefix tests are case-sensitive Python string checks. That can lead to treating a branch-like ref as “other namespace” and skipping it, turning a would-deny into allow. _(fix: Normalize candidate prefixes to lower-case before testing (`candidate_l = candidate.lower()`), and strip exactly `refs/heads/` in a case-insensitive way. Add tests for `refs/Heads/master` and `REFS/heads/Master`. If you prefer strictness, explicitly deny when a ref looks like a heads ref with mismatched casing rather than silently skipping.)_
- `hooks/git_policy.py:311-352` **[low] Primary-checkout predicate assumes fully-formed state; missing key could deny** _(confidence 0.41)_
  Probe-failure invariant claims “no probe failure can ever produce a deny,” but the path through primary_checkout uses only cached `state.get('linked_worktree') is False` to deny. If `state()` returns a partially-populated dict where linked_worktree defaults to False due to a truncated or malformed `rev-parse` output (lines <3 guarded earlier, but future changes or exotic git versions could surface `linked_worktree` missing and default False later), `_fires('primary_checkout')` will wrongly deny. Today `_probe` returns None on too-few lines, but this safety is brittle: any refactor that adds a default dict literal for `state` reintroduces the hazard. Tests do not cover a malformed `state` dict where keys are missing but non‑None. _(fix: Defensively check key presence: require `state.get('linked_worktree') is not None` and only deny when it is explicitly False. Otherwise return None. Add a unit test simulating a probe runner that returns a line count mismatch that still yields a dict missing `linked_worktree` to ensure no deny is produced.)_
- `hooks/git_policy.py:478-420` **[medium] Inconsistent scope: default-branch deny fires for mapped pushes that policy says are out-of-scope** _(confidence 0.50)_
  Set-based widening from “last refspec” to “any refspec” changed the false-positive profile for default-branch detection: `git push origin feature/x master` (mistyped order) now denies even if the user intended only feature. This is correct for safety, but the classification still happens when the push is outside the policy’s scope in other ways. Specifically, `_pushes_elsewhere` allows `git push origin <current-branch>` but treats any `src:dst` as outside scope. Yet `_push_targets` will still parse and include the `dst` for default-branch deny even when `_pushes_elsewhere` would have declared it out-of-scope for policy evaluation. If an entry combines predicates, an earlier ‘shared_history’ pass could be skipped by `only_flags`, but `default_branch_target` still fires on a mapped push that policy text says should fall back to ask. This is an internal inconsistency: out-of-scope pushes should not be evaluated for default-branch deny. _(fix: Gate default-branch evaluation on `_pushes_elsewhere(argv, state['branch'])` being False. If the push names a different ref than current, skip deny to match the documented fallback-to-ask behavior. Add tests for `git push origin HEAD:master` and `git push origin master:feature/x` asserting no default-branch deny via this path.)_

</details>

debedb:
Second pass. **One finding was real and is fixed, one led to dead code being
removed, three do not hold.**

The real one is the good kind: it is a regression the *previous* round's fix
introduced.

## Fixed: `--tags` was suppressing an explicit default-branch refspec

Last round I made `--all`, `--mirror` and `--tags` all "unresolvable". They
are not the same kind of thing, and lumping them together broke the guard:

```
git push origin master --tags   ->  None   (no deny)   <-- regression
git push origin master          ->  {'master'}  (deny)
```

`--tags` pushes tags *in addition to* whatever refspecs are given, so the
explicit `master` sitting next to it still targets the default branch. Now:

| command | targets | result |
| --- | --- | --- |
| `git push origin master --tags` | `{master}` | **DENY** |
| `git push --tags origin master` | `{master}` | **DENY** |
| `git push --tags` | none | ask (tags only, no branch) |
| `git push --all origin` | every branch | **DENY** |
| `git push --mirror origin` | every branch | **DENY** |

`--all` and `--mirror` push every branch, so the default branch is in that set
by definition — they resolve, they are not unknown. Checked the corpus before
widening: all three flags appear **zero times** in 12 days, so the stricter
reading costs no measured friction. 695 tests pass.

## Removed: `_pushes_elsewhere` was dead

Finding #4 reasoned about `_pushes_elsewhere` gating the default-branch
evaluation. It does not gate anything — the allow layer was its only caller
and the port left it orphaned. The finding's conclusion was wrong, but it
prompted an AST sweep for private functions with no in-file caller, which
found it. Deleted.

## Does not hold: empty-set vs None conflation

Both states produce no-deny today, which is correct for both. The concrete
suggestion — `if targets is None:` instead of `if not targets:` — is
behaviourally identical here: an empty set falls through to
`default_branch in targets`, which is `False`, and returns `None` anyway. A
readability preference, not a defect.

## Does not hold: case-sensitive `refs/heads` normalisation

The claim is that `refs/Heads/master` would be skipped, turning a would-deny
into an allow. Git ref namespaces **are** case-sensitive: `refs/Heads/master`
is a different ref from `refs/heads/master` and pushing it does not touch the
default branch. Skipping it is correct.

The recommended fix — lowercase before comparing — would introduce a **false
deny** on a push that does not target the default branch. This is the one
finding whose remedy would have made the code worse.

## Does not hold: `primary_checkout` could deny on a malformed state dict

The recommendation is to "require `state.get('linked_worktree') is not None`
and only deny when it is explicitly False." That is what the code already
does:

```python
if state.get("linked_worktree") is False:
```

A missing key yields `None`, and `None is False` is `False`, so it returns no
opinion. The recommendation was already implemented in the line the finding
quotes.

## On the re-review

Zero findings cleared the 0.6 confidence floor; the one that mattered came in
at 0.62 inline. Passing `--focus` with the settled findings named explicitly
("do not re-report X, Y, Z") did keep it off the previously-rebutted ground —
worth recording for the `codex-adversarial-pr-review` skill, alongside the
earlier note that `--focus` is what gets past the degenerate plan-only shape.

Net across both rounds: two real miss-denies found, one of them caused by
fixing the other.

debedb (commented):
[codex] Review r1 — summary and follow‑ups

Overall direction looks solid. The deny-side policy is narrowly scoped, the “no probe failure → no DENY” invariant is called out in prose, and the refspec parser now counts any destination that matches the default branch. The added tests are a strong signal. Two non‑blocking asks to lock behavior and one nit for clarity:

- blocking?: Verify end‑to‑end (ThroughTheClassifier) that probe failures never yield DECISION_DENY — e.g., TimeoutExpired, OSError, non‑zero exits, and empty/garbled outputs. If already covered by the nine invariant tests mentioned, just point me to the names; otherwise please add explicit boundary tests.
- non‑blocking: Ensure mixed refspec ordering is permanently guarded (both  and  deny when default branch is present); I see unit coverage for multi‑refspecs — a regression test here would future‑proof the “last‑only” bug.
- non‑blocking: Mirror the tri‑state authorship semantics (SOLO/SHARED/UNDETERMINED → fall through to pre‑classified unsafe→ask) in docstrings/comments near  so the rationale travels with the code.

If the boundary tests already exist and the refspec cases are asserted in both orders, this is LGTM from me.

Status: PR #122 @ HEAD %HEAD7%, base %BASE%, state %STATE%.

debedb:
[codex] Review r1 — summary and follow-ups

Overall direction looks solid. Two asks plus a clarity nit:

- blocking: Add end-to-end tests at the ThroughTheClassifier boundary that prove probe failures (timeout, OSError, non-zero exit, empty/garbled output) never produce DECISION_DENY.
- non-blocking: Add a regression test asserting mixed refspec ordering denies when the default branch is present, regardless of order; this future-proofs the previous last-refspec bug.
- non-blocking: Mirror the tri-state authorship semantics (SOLO/SHARED/UNDETERMINED -> falls through to unsafe->ask) in code comments near _maybe_deny_by_policy so the invariant travels with the code.

If these are already covered, please point me to the test names; otherwise, after landing them I’m good to approve.

debedb:
All three landed in `2d98075`. Taking the blocking one first, since you were
right and the honest answer to "are these already covered" is **no**.

## blocking: end-to-end probe-failure tests

They did not exist. The nine `FailedProbeNeverDenies` tests assert the
invariant against `GitDenyPolicy.evaluate()` directly — which is the layer
that **cannot see the wiring**. That distinction is not theoretical here:
three separate wiring bugs (an unimported `DECISION_DENY` in two modules, and
a `cd` hook that never landed) passed every one of those nine while the hook
crashed outright. So your instinct matches the exact failure this file had
already suffered.

New class `ProbeFailureNeverDeniesThroughTheClassifier`, driving a full
`classify()` call with the failure injected at the probe:

| test | failure injected |
| --- | --- |
| `test_state_probe_fails` | `rev-parse` returns nothing |
| `test_user_email_unset` | `git config user.email` unset |
| `test_unreadable_history` | `git log` returns nothing |
| `test_garbled_probe_output` | truncated multi-line output |
| `test_empty_probe_output` | empty string from every probe |

Every one asserts `unsafe` — not `deny`, and not silence. The static rules
already flagged `git push` as mutating, so a failed probe leaves that
standing and the operator is still asked.

## The gap you did not name, which was worse

Every test in the file injected a runner that **already returned `None`**.
Nothing asserted that the real `_run_git` converts a `TimeoutExpired`, an
`OSError` or a non-zero exit *into* that `None` — and that conversion is the
step the entire fail-open invariant rests on. You asked for `TimeoutExpired`
and `OSError` by name, which is what surfaced it.

New class `RunGitConvertsFailuresToNone`, replacing `subprocess.run`:
`test_timeout`, `test_oserror`, `test_nonzero_exit`, `test_success_returns_stdout`.

## Mutation-checked, not assumed

A test that passes for the wrong reason is worse than no test, so both
invariants were verified by reintroducing the bug:

| mutation | result |
| --- | --- |
| #82's original collapse — `UNDETERMINED` read as `SHARED` | **5 failures** |
| remove `DECISION_DENY` from the `grammar_classifier` import | **4 errors** |

The second is the one worth noting: that is the precise wiring bug that
previously passed **every** unit test in this file. It now errors. Green
before each mutation and after reverting it.

## non-blocking: refspec ordering, both directions

`test_every_refspec_counts_not_just_the_last` now asserts
`origin master feature/x` **and** `origin feature/x master`, and
`ThroughTheClassifier.test_mixed_refspec_order_denies_either_way` does the
same through a full classify. Your reasoning is why: a single-order test
passes for the wrong reason whenever the default branch happens to be last,
which is the exact shape of the bug it guards.

## non-blocking: tri-state semantics in the comments

`_maybe_deny_by_policy` now carries the table and the reason:

```
SOLO          -> no deny; `unsafe` stands, operator is asked
UNDETERMINED  -> no deny; `unsafe` stands, operator is asked
SHARED        -> deny
```

plus why `UNDETERMINED` deliberately shares a bucket with `SOLO` rather than
being a pessimistic default, and pointers to both test classes.

46 tests in this file, up from 28. 705 in the suite, all passing.

debedb:
[codex] Review r2 — follow-up

No new commits since my last comment; the previous asks still stand:

- blocking: Add end-to-end tests at the ThroughTheClassifier boundary that prove probe failures (timeout, OSError, non-zero exit, empty/garbled output) never produce DECISION_DENY.
- non-blocking: Add a regression test asserting mixed refspec ordering denies when the default branch is present, regardless of order.
- non-blocking: Mirror the tri-state authorship semantics in code comments near _maybe_deny_by_policy.

Once the boundary test is in (or you can point me to the exact test names if already present), I’m good to approve.

debedb:
Audited all three rounds against the code rather than against my own replies.
**Seven collapsed findings, two inline. Two of them I had answered with
neither a fix nor a rebuttal** — closed in `77f0a96`, and one of them turned
out to be hiding a real bug.

## The complete ledger

| # | round | finding | disposition |
| --- | --- | --- | --- |
| inline | r1 | refspec parsing, last-only | **fixed** `af01275` |
| inline | r2 | bulk flag suppresses explicit target | **fixed** `5befa4c` |
| 1 | r1 | `ProbeResult` sentinel; scattered None checks | rebutted — `TimeoutExpired` subclasses `SubprocessError` |
| 1b | r1 | *fuzz runner: empty / partial output* | **now covered** — `test_garbled_probe_output`, `test_empty_probe_output` |
| 1c | r1 | *gate `stderr=DEVNULL` for observability* | **was unanswered — now fixed** |
| 2 | r1 | `cd`/`-C` hardening (a)(b)(c) | rebutted — all fail-open |
| 2d | r1 | *brace expansion in `cd`* | **was unanswered — now fixed** |
| 3 | r1 | segment-scoped DENY precedence | declined by design |
| 4 | r2 | empty-set vs `None` | rebutted — behaviourally identical |
| 5 | r2 | case-sensitive `refs/heads` | rebutted — remedy would add a false deny |
| 6 | r2 | `primary_checkout` missing key | rebutted — `is False` already does it |
| 7 | r2 | mapped pushes via `_pushes_elsewhere` | moot — function was dead, removed |
| — | r1 (today) | three asks | **done** `2d98075` |

## The two that were genuinely open

**`stderr=DEVNULL` (1c).** Discarding by default is right — a failed probe is
a normal outcome and git chatter does not belong in the operator's session.
Having *no escape hatch* is the wrong default in a tool whose failure mode is
silently ceasing to guard (#123). `YOLT_POLICY_DEBUG=1` now lets it through,
read at call time.

**Brace expansion in `cd` (2d).** Denies nothing either way — an unexpanded
path does not exist, so the probe fails. But the unresolvable set should
state intent, not list the cases someone happened to think of. `{`, `}`, `?`
join `$`, backtick, `*`.

## Writing that test found a real bug, inherited from #82

```python
args = [a for a in argv[1:] if not a.startswith("-")]
...
if target == "-" or ...          # unreachable
```

A bare `-` is filtered out as a flag **before** the `-` test can see it. So
`cd -` falls through to the empty-args branch and resolves to `$HOME`. It
means the *previous* directory, which no static walk can know, so the correct
answer is unresolvable.

Verbatim from `feature/git-worktree-allow`, same structure, same position —
latent since #82. Harmless there: a wrong directory cost a grant that did not
happen. On the deny side it judges the wrong repository. The check now runs
before flags are filtered.

```
cd -         -> None   (was: $HOME)
cd {a,b}     -> None   (was: literal path)
cd $X        -> None
cd /tmp/real -> /tmp/real
```

707 tests. @reviewer — r1's three asks are in `2d98075` with the test names
in the comment above; this covers the older rounds you may not have seen.

debedb:
[codex] Approval — r3

Thanks for landing:
- ThroughTheClassifier boundary tests proving “no probe failure → no DENY”.
- Mixed refspec ordering and --tags alongside explicit branch refspec denial tests.

LGTM.

PR #126, opened 2026-09-14
Let a consumer that is not the terminal opt out of its allow-lists (--no-user-allow)

### LINKED ISSUE #125: Let a non-terminal consumer opt out of Claude Code's allow-lists (--no-user-allow)
`run_cli()` loads `Bash()` allow patterns from `~/.claude/settings.json`,
`<cwd>/.claude/settings.json` and `<cwd>/.claude/settings.local.json`
(`grammar_classifier.py:517-519`), with no way to say no. For the PreToolUse
hook that is exactly right -- those files are the permissions of the human at
that terminal.

For any other consumer it is a confused deputy. voitta-ai/shmobster runs a Slack
agent whose commands go through this classifier: a command auto-runs when the
verdict is `safe`, and everything else parks for a human's approval. So the
Slack agent's auto-run set is, silently, whatever the operator once allowed
themselves in a terminal. Measured on that deployment: **123 patterns**, which
today promote `gh pr merge <n> --squash`, `gh api -X DELETE <repo>`,
`git push origin feature/*` and `codex exec "<anything>"` to `safe` -- all
mutating, none of them ever seen by a card.

This repo's own test suite has the same problem from the same cause: running
`python3 -m unittest discover tests` from a clone whose `.claude/settings.local.json`
carries 52 patterns fails 15 tests; from a worktree that lacks that untracked
file, 11. The suite's verdict depends on whose machine it runs on
(`test_git_tag_delete_is_unsafe`, `test_git_push_reason_includes_allow_hint`
and the `find`/`gh api` regressions are the visible ones).

## Ask

A `--no-user-allow` flag on the CLI that skips settings discovery, so a consumer
that is not an interactive terminal can get the rules' own verdict. Default
unchanged -- the hook keeps inheriting, because there the inheritance is the
feature.

Worth pairing it with the count of patterns actually in play in the JSON output,
so a consumer can log the size of its auto-run surface at startup rather than
guess at it.

Downstream: voitta-ai/shmobster#148.

debedb:
Closes #125.

`run_cli()` reads `Bash()` allow patterns from `~/.claude/settings.json` and the
project's `.claude/settings*.json` with no way to decline. For the PreToolUse
hook that is the feature -- those files are the permissions of the human at that
terminal. For a consumer that is *not* that terminal it is a confused deputy:
voitta-ai/shmobster's Slack agent auto-runs whatever comes back `safe`, so its
auto-run set is silently whatever the operator once allowed themselves.

Measured on that deployment -- 123 patterns, and today:

    gh pr merge 1 --squash    -> safe, "matches user allow pattern"
    codex exec "<anything>"   -> safe, "matches user allow pattern"
    gh api -X DELETE <repo>   -> safe, "matches user allow pattern"

all mutating, none ever seen by an approval card.

## The change

    $ python3 hooks/grammar_classifier.py 'gh pr merge 1 --squash'
    {"decision": "safe", "reason": "matches user allow pattern 'gh pr merge*'", "allow_patterns": 106}

    $ python3 hooks/grammar_classifier.py --no-user-allow 'gh pr merge 1 --squash'
    {"decision": "unsafe", "reason": "gh pr merge: mutating", "allow_patterns": 0}

    $ python3 hooks/grammar_classifier.py --no-user-allow 'cat x'
    {"decision": "safe", "reason": "cat: read-only", "allow_patterns": 0}

Default unchanged; only the consumer knows which it is. `allow_patterns` is new
in the output -- the same command classifies differently under a different
settings file, so the count is part of what the verdict means, and it lets a
consumer log the size of its auto-run surface at startup instead of guessing.
Existing consumers read `decision`/`reason` and are unaffected by an added key.

## Tests

Three new cases in `TestClassifierCLI`: inheritance still upgrades by default,
`--no-user-allow` drops it, and a read-only command is unaffected either way.
The class now writes its own settings file under a temp `HOME` and runs from a
directory with no project `.claude/`, which is the first time these CLI tests do
not depend on the developer's own permissions.

The rest of the suite still does, and the delta is worth recording: from a full
clone 15 tests fail; from a worktree lacking the untracked
`.claude/settings.local.json` (52 patterns, including `Bash(git --no-pager
push)` and `Bash(git checkout *)`), 11 -- the four that differ are the `git
push` / `git tag` ones. Same defect, one layer up. Pre-existing on master, noted
in #125, not touched here. No test fails on this branch that passes on master.

MINOR: new flag and a new output key, no change to any existing verdict.
Manifests bumped to 1.2.0 in both.

https://claude.ai/code/session_01X2TV3wYGyX3obgQDXg4jBz


debedb:
Merged. Two notes from the session working on `feature/auto-mode-realignment`,
recorded here because they change how this flag should be read later.

**The umbrella deletes what this flag turns off.** Phase 1 (#104) removes the
allow-pattern path wholesale -- `load_allow_patterns`, `match_allow_patterns`
and the `allow_patterns=` plumbing do not exist on that branch, because YOLT
stops emitting `permissionDecision: allow` at all. So when 2.0.0 lands,
`--no-user-allow` will have nothing left to disable. That is not a regression
and should not be read as one: the exposure it closes is live on master today,
and 2.0.0 is not close.

**Ask for the reconciliation list** (already noted on that side): keep
`--no-user-allow` *accepted as a no-op* in 2.0.0 rather than removing it.
Consumers pass it unconditionally -- shmobster does, and treats an
unrecognized flag as a fail-closed startup warning -- so an accepted no-op
costs one line upstream and saves every consumer a version gate.

**Same shape as #98.** A Slack agent auto-running whatever classifies `safe`,
with the auto-run set silently inherited from someone else's terminal, is
structurally what #98 found inside YOLT: approvals granted under auto mode that
nothing could see. One host, different building.

**The suite's dependence on the developer's settings is #75**, open and
unclaimed -- not a side effect of this PR. The workaround, for anyone hitting
the 15-vs-11 failure delta this PR's body describes:

    CLEAN=$(mktemp -d); HOME="$CLEAN" python3 -m unittest discover -q tests

705 pass that way.

PR #127, opened 2026-09-14
Refuse writes that steer the agent, not just the one that disables the hook

### LINKED ISSUE #121: Agent-steering writes are classified safe: unsafe_write_targets covers settings.json but not skills, agents, hooks, commands, or MCP config
## What

`rules/shell.json` lists `~/.claude/*` in **`safe_write_targets`**. Its
`unsafe_write_targets` covers exactly two paths under that tree:

```
~/.claude/settings.json
~/.claude/settings.local.json
```

Everything else under `~/.claude` is therefore classified **safe** — which
post-Phase-1 means the hook exits silently and never renders an opinion.

That includes:

```
~/.claude/skills/**       loaded into the model's context automatically
~/.claude/commands/**     slash commands the operator invokes by name
~/.claude/agents/**       subagent definitions, including their tool grants
~/.claude/hooks/**        code the host executes on tool calls
~/.claude/plugins/**      the extracted plugin cache, incl. this hook's own copy
~/.claude/CLAUDE.md       standing instructions
.mcp.json                 (project-level; in neither list)
```

## Why it matters

The comment in the rules file already states the principle, for the one path
it applies it to:

> `~/.claude/settings.json` is unsafe even though `~/.claude/*` is a
> safe-write target, **because settings.json can disable this hook**.

That reasoning generalizes and was not generalized. A skill file steers the
model's behavior on every future turn; an agent definition carries tool
grants; `~/.claude/hooks/**` is executed by the host. A write to any of them
is the self-modification shape, and it is the shape #100 names as the *first*
entry on the non-delegable list.

There is also an asymmetry that reads backwards for an agent-safety tool:
`~/.bashrc` and `~/.zshrc` are `unsafe` because they steer a **shell**, while
`~/.claude/skills/*` is `safe` though it steers the **agent**.

## Evidence

Measured on the 12-day dogfood corpus with `scripts/replay_unsafe.py`, over
the 3,846 intact commands:

- **184 commands touch agent-steering paths.**
- Decision mix: **158 `safe`**, 16 `unsafe`, 10 `unknown`.
- The 16 `unsafe` are unsafe *incidentally* — they contain `cp`, `mkdir`,
  `git pull` — not because of where they write. Examples in the corpus
  include `cp <src>/SKILL.md ~/.claude/skills/<name>/SKILL.md` and
  `mkdir -p ~/.claude/skills/<name>/.state && date +%F > ...`.

So the incidental coverage that exists today comes from the general-purpose
`cp` / `mkdir` rules — which is exactly what Phase 3 (#100) proposes to
delete. Without this gap closed first, Phase 3 removes the accidental
coverage and leaves the intended coverage still unwritten.

## Proposed

Move the steering subtree out of `safe_write_targets` and enumerate it in
`unsafe_write_targets`:

```
~/.claude/skills/*
~/.claude/commands/*
~/.claude/agents/*
~/.claude/hooks/*
~/.claude/plugins/*
~/.claude/CLAUDE.md
.mcp.json
```

Keep `~/.claude/*` safe for the genuinely incidental writes it exists for
(logs, caches, `~/.claude/yolt/*` state), since `unsafe_write_targets` is
checked first and so the specific entries win.

Worth confirming against the corpus before merging: the 158 currently-`safe`
commands are dominated by *reads* of those paths and writes to skill `.state`
files, so the added prompt volume should be small. `scripts/replay_unsafe.py
--diff` will say exactly.

## Note on scope

This is a rules change with user-visible effect, so it is a **minor** bump,
and it belongs on the umbrella rather than master if it lands as part of
#100's non-delegable list. Filing separately because it is a gap in shipped
v1.1.0 behavior and is worth deciding on its own merits, not only as a Phase 3
subtask.

debedb:
Closes #121.

## The hole

`rules/shell.json` listed `~/.claude/*` in **`safe_write_targets`**, and
named only two paths beneath it in `unsafe_write_targets`:

```
~/.claude/settings.json
~/.claude/settings.local.json
```

Everything else under that tree classified **safe** — which post-Phase-1
means the hook exits silently and renders no opinion at all:

```
~/.claude/skills/**     read into the model's context on every later turn
~/.claude/agents/**     subagent definitions, including their tool grants
~/.claude/commands/**   slash commands the operator invokes by name
~/.claude/hooks/**      code the host executes on tool calls
~/.claude/plugins/**    the extracted plugin cache, incl. this hook's own copy
~/.claude/CLAUDE.md     standing instructions
.mcp.json               servers the agent will trust
```

The file already stated the principle for the one path it applied it to:

> `~/.claude/settings.json` is unsafe even though `~/.claude/*` is a
> safe-write target, **because settings.json can disable this hook**.

That was never generalised. The asymmetry it left is the cleanest way to see
it: **`~/.bashrc` was unsafe because it steers a shell, while
`~/.claude/skills/*` was safe though it steers the agent.**

Codex paths are included on the same grounds now that a `.codex-plugin`
manifest ships: `config.toml`, `hooks.json`, `AGENTS.md`, `skills/`.

## Measured before it was written

Replayed against the 12-day corpus, 3,846 intact commands:

```
newly unsafe (added prompts): 0
lost unsafe (regressions):    0
```

**Zero is the honest number, and it is not an argument that the change is
idle.** The corpus's steering-path writes were already `unsafe`
*incidentally* — via the general-purpose `cp` and `mkdir` rules, never
because of where they wrote. Those are exactly what Phase 3 (#100) proposes
to delete. Landing this first is what keeps the coverage when the accident
goes away; land it after and there is a window with neither.

Verified separately that the rule actually fires, since a zero delta could
equally mean a dead rule:

| command | before | after |
| --- | --- | --- |
| `echo x > ~/.claude/skills/evil/SKILL.md` | safe | **unsafe** |
| `echo x > ~/.claude/CLAUDE.md` | safe | **unsafe** |
| `echo x > ~/.codex/config.toml` | unknown | **unsafe** |
| `echo x > .mcp.json` | unknown | **unsafe** |
| `echo x > ~/.claude/yolt/cache.json` | safe | safe |
| `cat ~/.claude/skills/x/SKILL.md` | safe | safe |

`~/.claude/*` stays a safe-write target for everything else beneath it —
logs, caches, `~/.claude/yolt/` state — because `unsafe_write_targets` is
checked first and the specific entries win.

## Tests come in pairs

Each steering path refused, **and** the ordinary path beside it untouched. A
guard that flipped everything under `~/.claude` would pass half this file and
be useless, so the controls are half the point.

Mutation-checked rather than assumed: reverting the list to master's fails
**10 of 19**, and the 9 that pass either way are exactly the controls. 693 in
the suite.

## Version

`1.2.0` -> `1.3.0`, both manifests. MINOR: a user-visible classification
change, no config key or format break.

## Not in scope

Writes reaching these paths through the Python analyzer (`python3 -c
"open(...)"`) or through the `Write` tool rather than Bash.
`unsafe_write_targets` covers redirects and the write-target arguments of
`tee`/`cp`/`mv`/`install`/`dd`/`find`. The `Write` tool path is Phase 2
(#99/#106), which widens the matcher past `Bash`.


debedb:
## Scope narrowed: this PR is now only the rules list

While probing this change for the adversarial review I found a more severe
bug and filed it as **#128**. The path-normalisation hardening I had written
here belongs to that fix, so I have reverted it out — this PR now touches no
hook code at all, only `rules/shell.json`, its tests, and the bump.

**#128 in one line:** `echo x > $HOME/.ssh/authorized_keys` classifies
`safe`, and on master `safe` emits `permissionDecision: allow` — so YOLT
actively grants it, while `~/.ssh/authorized_keys` asks. Writing `$HOME`
instead of `~` defeats every entry in `unsafe_write_targets`, including
`~/.claude/settings.json`, which is on the list *because it can disable this
hook*. Root cause is `_redirect_write_target` extracting only `word` and
`string` nodes while tree-sitter parses `$HOME/...` as a `concatenation`, so
the redirect is dropped from consideration entirely rather than merely
unmatched.

#128 is shipped and grants; this one is a narrower hole that only exits
silently. Landing #128 first, then rebasing this.

**Note on the version:** this PR currently claims `1.3.0`. If #128 lands
first it takes that number and this one moves to `1.4.0` — I will re-bump at
merge time rather than now, per the rule that the bundle number is taken when
it lands, not when the branch is built.

## On the adversarial review

Its top finding (0.77, `high`) claimed `fnmatch`'s `*` is shallow so nested
steering files evade the globs, and that the tests never exercise nesting.
Both halves are false: Python's `fnmatch` matches `/` with `*` (unlike
`glob`), nested paths classify `unsafe` at any depth, and
`test_plugin_cache` already asserts a five-component path.

Recording it because the useful part is what happened next — probing the
*category* it named, rather than the claim, is what surfaced #128.

PR #129, opened 2026-09-14
Read a write target spelled with a variable, and never mistake one for absent

### LINKED ISSUE #128: Writing $HOME instead of ~ bypasses every unsafe_write_target, and master grants it
## Every entry in `unsafe_write_targets` is bypassed by spelling the path with a variable

Shipped behaviour on master (v1.2.0), through the real hook:

```
ASK     echo x > ~/.ssh/authorized_keys
ALLOW   echo x > $HOME/.ssh/authorized_keys
```

Not "fails to ask" — **actively grants**. On master a `safe` decision emits
`permissionDecision: allow`, which short-circuits auto mode, so the operator
is never prompted by anything.

It applies to the whole list, not one entry:

```
safe   echo x > $HOME/.ssh/authorized_keys
safe   echo x > $HOME/.aws/credentials
safe   echo x > $HOME/.bashrc
safe   echo x > $HOME/.claude/settings.json
safe   echo x > ${HOME}/.claude/settings.json
```

`~/.claude/settings.json` is on the deny list specifically **because it can
disable this hook**. `$HOME/.claude/settings.json` is granted.

## Root cause

`_redirect_write_target` in `hooks/grammar_classifier.py` extracts only two
node types:

```python
elif c.type == "word" and target is None:
    target = self._slice(c, src)
elif c.type == "string" and target is None:
    target = self._reconstruct_string(c, src)
```

tree-sitter parses the two spellings differently:

```
echo x > ~/.ssh/authorized_keys        child type='word'            '~/.ssh/authorized_keys'
echo x > $HOME/.ssh/authorized_keys    child type='concatenation'   '$HOME/.ssh/authorized_keys'
```

A `concatenation` matches neither branch, so `target` stays `None` and the
function returns `None`. The caller reads that as *"this redirect is not a
write"* — so the redirect is not merely unmatched against the deny list, it
is **dropped from consideration entirely**. The command is then classified on
`echo` alone and comes out `safe`.

So the failure is not a glob that missed. The write target never reaches the
glob.

## Two fixes, and the second matters more

**1. Extract `concatenation` targets.** Slice the surface text so
`$HOME/...` becomes a literal string, then expand it. `_expand_home`
currently handles only a `~/` prefix and needs `$HOME/` and `${HOME}/` too,
plus `normpath` so `~/.claude//skills/x` cannot evade a glob by doubling a
slash (also currently a bypass, same root).

**2. A write target that cannot be extracted must not read as "no write".**
This is the structural half. Any future node type nobody anticipated — a
command substitution, an array expansion, a form a later tree-sitter release
parses differently — silently becomes a grant today. An unextractable write
redirect should classify `unknown` at worst, never fall through to `safe`.

Fix 1 without fix 2 closes the spelling that is known and leaves the shape
that produced it.

## Severity

Worth stating plainly because the repo has been here before. This is #98's
finding in a new place: **YOLT granting approvals that nothing else can
see.** The deny list looks enforced, reports itself as enforced, and is one
keystroke from optional. Anything that can put a command on the line —
including content the agent read from somewhere else — gets an active grant
on ssh keys, cloud credentials, shell startup files, and the hook's own
configuration.

Found while probing #121/#127. Unrelated to that change: reproduced on
unmodified master.

## Note for the fix

`~/.claude/skills/../../.bashrc` already resolves onto the `~/.bashrc` entry
once `normpath` is applied, so path traversal is covered by the same fix
rather than needing its own.

debedb:
Closes #128.

## What was shipped

Through the real hook, on master (v1.2.0):

```
ASK     echo x > ~/.ssh/authorized_keys
ALLOW   echo x > $HOME/.ssh/authorized_keys
```

**Not "fails to ask" — actively grants.** A `safe` decision on master emits
`permissionDecision: allow`, which short-circuits auto mode, so nothing else
prompts either. It applied to every entry in `unsafe_write_targets`:

```
safe   echo x > $HOME/.aws/credentials
safe   echo x > $HOME/.bashrc
safe   echo x > $HOME/.claude/settings.json      <- on the list because it can disable this hook
```

## Root cause

`_redirect_write_target` read only `word` and `string` nodes. tree-sitter
parses the two spellings differently:

```
> ~/.ssh/authorized_keys        child type='word'           '~/.ssh/authorized_keys'
> $HOME/.ssh/authorized_keys    child type='concatenation'  '$HOME/.ssh/authorized_keys'
```

A `concatenation` matched neither branch, so `target` stayed `None` — and the
caller reads `None` as *"this redirect is not a write."* The redirect was
therefore **dropped from consideration entirely** rather than merely going
unmatched, and the line was classified on `echo` alone.

The deny list never failed. The target never reached it.

## Three parts; the third is the one that matters in a year

**1. Read the target.** `concatenation` and the expansion node types are now
taken as surface text, so `$HOME/...` arrives at the matcher as a literal.

**2. Normalise it.** `_expand_home` resolves `$HOME/` and `${HOME}/` beside
`~/`, and runs `normpath` — which also closes `~/.claude//settings.json` (a
second bypass, same root) and makes `~/.claude/../.bashrc` land on the
`~/.bashrc` entry instead of escaping both.

The classifier now calls the rule layer's function instead of keeping its
own copy. The two had identical bodies and identical blind spots, so this bug
was open in both places and fixing either alone would have left the other.

**3. Unreadable is not absent.** A write redirect whose target cannot be read
returns a sentinel distinct from `None` and downgrades the statement to
`unknown`. That conflation is what made this a grant rather than a prompt,
and any node shape nobody has anticipated yet lands in exactly the same
place. Friction, never silence.

Part 3 without parts 1-2 would have left the known spelling open; parts 1-2
without part 3 would have closed this spelling and left the shape that
produced it.

## Verified through the hook, not just the classifier

```
ASK      echo x > $HOME/.ssh/authorized_keys
ASK      echo x > $HOME/.aws/credentials
ASK      echo x > ~/.ssh/authorized_keys
ALLOW    echo x > /tmp/ok
```

## Mutation-checked in three parts

One mutation per part, because a single revert only proves the import:

| mutation | failures |
| --- | ---: |
| drop the `concatenation` branch | 6 |
| drop `$HOME` from the normaliser | 6 |
| let an unreadable target fall through again | 1 (the structural test) |

690 in the suite.

## Version

`1.2.0` -> `1.3.0`, both manifests. MINOR: a user-visible classification
change, no config key or format break.

## Relationship to #127

Found while probing #127 for its adversarial review, but reproduced on
unmodified master and independent of it. #127 is narrower — steering paths
that only exit silently — and has been rescoped to the rules list alone. It
will re-bump to `1.4.0` after this lands, per taking the version at merge
time rather than branch time.


debedb (commented):

Do not ship. The fix closes bare `$HOME` concatenations in redirections but leaves command/backtick substitution forms and argv write‑args (tee/cp/mv/install/dd/find) able to evade protected‑path detection. UNEXTRACTABLE downgrades hide clear deny cases, and `normpath` may alter prior verdicts. Tests don’t cover these high‑risk spellings or benign variable‑safe paths, so security and friction regressions are both plausible.

### Next steps
- Add unit tests for substitution/backtick write targets across redirects and argv write‑args (tee/cp/mv/install/dd/find).
- Tighten `_redirect_write_target` and `check_unsafe_flags` to treat tokens containing substitution nodes as non‑bypassable for deny‑list checks, upgrading to `unsafe` when protected‑path suffixes are textually present.
- Constrain `_expand_home` normalization scope and add tests for relative and mixed HOME/non‑HOME normalization to prevent silent verdict shifts.
- Add UNEXTRACTABLE heuristic protected‑path detection on the raw slice to avoid downgrades to `unknown` for obvious deny cases.
- Add safe‑path variable expansion tests to guard against friction‑only regressions.

### Findings outside the diff
_(could not be attached inline; lines are not part of this PR's diff)_
- `hooks/grammar_classifier.py:21-146` **[high] Protected-path bypass via command/backtick substitution in redirect target** _(confidence 0.78)_
  Command-substitution and backtick spellings of HOME-based protected files likely still evade the deny list because redirect-target extraction treats command/process/arithmetic substitutions inside quoted strings as the `SUBSTITUTION_PLACEHOLDER` and never reconstructs their textual output. For unquoted forms, `_redirect_write_target` only accepts `word`, `string`, and the `_REDIR_TARGET_NODES` tuple of `(concatenation, simple_expansion, expansion)`. Tree‑sitter emits `$(...)` and backticks as `command_substitution`/`process_substitution`, which are not in `_REDIR_TARGET_NODES`. In `_node_text` (used for argv reconstruction and thus for quoted targets), those nodes return `SUBSTITUTION_PLACEHOLDER` as well. Net effect: `echo x > $(echo $HOME)/.ssh/authorized_keys`, ``echo x > `echo $HOME`/.ssh/authorized_keys``, and `echo x > "$(echo $HOME)/.ssh/authorized_keys"` yield an UNEXTRACTABLE or placeholder target that fails to match the unsafe patterns, and the redirection path downgrades to `unknown` instead of `unsafe`. That is friction instead of a hard deny on a protected path write, leaving a high‑value path open to a one‑token bypass if allow patterns or user workflows later auto‑approve unknowns. The new tests do not cover these spellings. _(fix: Extend redirect-target extraction to treat `command_substitution` and `process_substitution` within redirection targets as UNEXTRACTABLE that is then checked first for protected-path semantics using literal surface text capture, or—preferably—treat any write redirect whose target contains substitution nodes as at least `unsafe` when the expanded path would match a deny glob if `$HOME` were substituted. Concretely: (1) include `command_substitution` and `process_substitution` in `_REDIR_TARGET_NODES` and have a branch that sets a `saw_substitution=True`; (2) when `saw_substitution`, return a sentinel like `UNEXPANDABLE_WITH_SUBSTITUTION` and in `_walk_redirected` classify to `unsafe` if the slice contains a `$HOME`/`~` pattern joined with protected suffixes (authorized_keys, .bashrc, etc.), else `unknown`; (3) add tests for `$(echo $HOME)/...` and backtick forms, both quoted and unquoted; (4) document that substitutions in write-targets are conservatively classified and cannot grant.)_
- `hooks/rule_classifier.py:66-74` **[high] Deny‑list gap for tee/cp/mv/install/dd/find write arguments with substitutions** _(confidence 0.75)_
  The identical concatenation blind spot likely persists on the other path into the deny list: write-target arguments parsed at the argv layer (tee, cp, mv, install, dd, find) in `check_unsafe_flags`. That function matches values and positionals only as raw tokens and via `_expand_home(value)`. It does not decompose tokens that are concatenations or include command substitutions; it relies entirely on what `grammar_classifier` passed into argv. Since `_argv_from_command` converts any `command_substitution`/`process_substitution` to `SUBSTITUTION_PLACEHOLDER`, a call like `tee $(echo $HOME)/.ssh/authorized_keys` or `cp src $(echo $HOME)/.bashrc` will never match the unsafe deny globs here and will fall back to the command’s `default` classification (often `unsafe` but without the specific protected‑path reason). Worse, for commands whose default is not `unsafe`, this becomes a silent downgrade. No tests cover these tee/cp/mv/install/dd/find argv cases with substitutions or quoted concatenations. _(fix: Harden `check_unsafe_flags`: when evaluating (a) `write_flag_value_targets`, (b) `write_value_prefix_targets`, and (c) `write_target_*_positional`, treat any token that equals the `SUBSTITUTION_PLACEHOLDER` or that contains it as an unexpandable write-target and conservatively classify the entire command at least `unknown` (and `unsafe` if a `$HOME`-expanded heuristic on the raw slice would land on a protected path). Alternatively, plumb a richer token structure from the grammar walker that marks tokens containing substitutions, and have this function treat them as non‑bypassable for deny‑list purposes. Add tests for `tee`/`cp`/`mv`/`install`/`dd of=`/`find -fprint` with `$(...)` and backticks around `$HOME`.)_

- `hooks/rule_classifier.py:35-66` **[medium] normpath may rewrite non‑HOME paths and shift prior verdicts** _(confidence 0.54)_
  `_expand_home` normalizes with `os.path.normpath` and preserves a trailing slash. While intended to close duplicate‑slash and `..` traversal spellings, it also rewrites relative paths like `./.bashrc` to `.bashrc` and collapses multiple slashes inside arbitrary patterns. If any safe/unsafe glob in rules relies on leading `./` or double‑slash semantics, this can change a prior `safe` verdict into `unsafe` or vice versa. The tests assert only a few happy cases. There is no coverage for relative dotpaths (`./.cache/x`), mid‑path double slashes on non‑HOME paths (`/tmp//ok`), or `..` that climbs above HOME before re‑entering a protected suffix (e.g., `/tmp/../$HOME/.ssh/authorized_keys`), which `normpath` may collapse in a way that makes raw‑versus‑expanded fnmatch disagree. _(fix: Constrain normalization to only HOME‑prefix paths and only to collapse duplicate slashes and `..` segments after a resolved HOME boundary. Avoid rewriting relative paths or non‑HOME absolute paths before deny/allow comparison. Add tests for: `echo > ./file`, `/tmp//ok`, `$HOME/../$HOME/.bashrc`, and `./.bashrc` to ensure verdicts are unchanged from pre‑change behavior except where explicitly intended.)_
- `tests/test_variable_write_targets.py:110-126` **[medium] Potential prompt‑only regressions for benign variable‑based safe writes** _(confidence 0.55)_
  Friction regressions: the change intentionally converts unreadable redirect targets into `unknown` instead of treating them as absent. That is correct for security but raises the risk of turning previously `safe` calls into prompts without a matching safety gain, especially when the write target is an ordinary safe path constructed via benign substitutions or array expansions (e.g., `> "$XDG_CACHE_HOME/app"`, `> "${arr[0]}/log"`). The tests include only `$HOME/.cache/thing` as a safe case; they do not cover XDG cache dirs, temporary directories pulled from variables, or array/substring expansions. Teams that depend on noninteractive flows could see unexpected prompts. _(fix: Broaden positive coverage to protect non‑dangerous, variable‑driven safe paths from unnecessary prompts. Add tests for `> "$XDG_CACHE_HOME/app"`, `> "${TMPDIR}/ok"`, `> "/tmp/${user}/x"`, and array/substring expansions like `> "${paths[0]}/ok"`. In the classifier, treat variables that expand under `_expand_home` to known safe prefixes as safe even when the raw target was reconstructed from concatenation, and avoid classifying such cases as unknown.)_

</details>

hooks/grammar_classifier.py
```diff
@@ -35,6 +36,15 @@
 
 _BASH_LANG = Language(_tsb.language())
 
+# Redirect-target node types that carry readable surface text. A
+# `concatenation` is the `$HOME/path` shape; the expansions are a bare
+# `$VAR` target, which resolves to no known-safe glob and so costs a prompt.
+_REDIR_TARGET_NODES = ("concatenation", "simple_expansion", "expansion")
+
+# A write redirect whose target could not be read. Distinct from None, which
```
debedb, line 44: **[high] UNEXTRACTABLE downgrade hides protected‑path writes under new node shapes** _(confidence 0.60)_

UNEXTRACTABLE sentinel handling in `_walk_redirected` only upgrades the decision to `unknown`. If the literal slice of the redirect contains a clearly protected HOME‑based suffix (e.g., `$HOME/.ssh/authorized_keys`) but extraction failed due to a new node shape, the classifier will not surface the strong protected‑path reason and may degrade to a generic friction prompt. Worse, heredocs and nested command substitutions bypass this path entirely: heredocs route only into Python inline analysis; command substitutions within redirect targets become placeholders and are not re‑classified in isolation. This creates a risk that future unparseable targets regress to `unknown` when they should be `unsafe`.

**Recommendation:** Augment UNEXTRACTABLE handling: when target is UNEXTRACTABLE, inspect the raw redirect slice and, if it contains `$HOME`/`${HOME}`/`~` joined to any deny‑listed basename or glob, classify `unsafe` with a protected‑path reason. Add explicit tests for unreadable targets that textually contain `$HOME/.ssh/authorized_keys` and `${HOME}/.bashrc` so they do not silently downgrade to unknown.

debedb:
Checked each claim against the tree. **Both named findings are refuted — but
measuring the category behind them found a real regression in this PR, now
fixed in `fb56935`.** Second time in two PRs that the finding was wrong and
its neighbourhood was not.

## The regression the review did not name, and its own question found

Finding 5 asked generically about friction regressions. So I replayed the
whole PR against the 3,846-command dogfood corpus, which surfaced something
worse than friction:

```
before: unsafe   terraform destroy: mutating
after:  unknown  writes to a file via redirection

terraform destroy -target=... -auto-approve > $SC/ekstf-destroy-run.log 2>&1
```

Two real lines. **A destructive command losing its prompt** — the one
direction that actually matters.

`_walk_redirected` returned as soon as a write target was not known-safe, so
the command itself was never classified. Latent before this PR: an unreadable
target used to be *dropped*, `write_targets` came back empty, and the command
got classified by accident. Reading the target correctly removed the
accident.

Aggregation already ranks unsafe above unknown, so the fix is to stop
returning early and let both decisions reach it.

Re-measured after the fix, same corpus:

```
6  safe -> unknown     [tightened]
3  unknown -> unsafe   [tightened]
   no loosening
```

The three that tightened are mutating verbs that had been masked this way all
along — two `git checkout` / `git pull` chains behind a redirect, and a write
into the Claude config directory behind another.

## Refuted: argv write-args (tee/cp/mv/install/dd/find) evade the deny list

```
unsafe   cp /tmp/x $HOME/.claude/settings.json
unsafe   tee $HOME/.ssh/authorized_keys < /dev/null
unsafe   mv /tmp/x $HOME/.bashrc
unsafe   install /tmp/x $HOME/.bashrc
```

Two independent reasons, either sufficient. That path reaches the deny list
through `_path_matches_target_list`, which calls `_expand_home` — **the
function this PR hardened** — so it got the `$HOME` fix for free. And those
verbs are already `unsafe` by verb, so the target never decided the outcome.

The finding assumed the argv path needed the same patch as the redirect path.
It needed the same *normaliser*, and now shares it.

## Refuted: command substitution and backtick forms still evade

```
unknown   echo x > $(echo $HOME)/.ssh/authorized_keys
unknown   echo x > `echo $HOME`/.ssh/authorized_keys
unknown   echo x > ~root/.ssh/authorized_keys
```

`unknown`, not `safe` — they do not grant. They cannot be `unsafe`: the
target is genuinely undecidable without executing the substitution. Landing
on `unknown` is exactly what part 3 of this PR guarantees. The recommendation
to upgrade them to `unsafe` on a textual suffix match would deny on a string
that may not be the path at all.

## The remaining claims

**`normpath` alters prior verdicts** — measured: 9 changes across 3,846
commands, all tightening, none loosening.

**Sentinel propagation** — `test_unreadable_target_downgrades_the_whole_statement`
forces the sentinel and asserts the statement downgrades. Mutation-checked:
letting it fall through again fails that test.

## Note on the review

Its verdict was "Do not ship," and it was right — for a reason it did not
identify. Both named findings are wrong; the generic question in finding 5 is
what prompted the corpus replay that caught a real destructive-command
regression.

Worth recording as a shape for `codex-adversarial-pr-review`: **a refuted
finding's category can still be load-bearing.** Rebutting the claim is not the
same as dismissing the neighbourhood, and measuring the neighbourhood anyway
is cheap.

694 tests.

hooks/grammar_classifier.py
```diff
@@ -167,12 +180,24 @@ def _walk_redirected(self, node, src, decisions, _depth):
...
                 seg, (DECISION_UNKNOWN, "writes to a file via redirection"),
             ))
-            return
+            # Deliberately NOT returning. An unknown redirect target must not
+            # mask the command's own verdict: `terraform destroy ... > $LOG`
+            # is `terraform destroy: mutating` first and an unclassifiable
+            # redirect second. Returning here dropped the destroy entirely,
+            # and aggregation already ranks unsafe above unknown, so falling
+            # through yields the stricter of the two rather than whichever
+            # was noticed first.
```
debedb: **[medium] Reason aggregation may double-count and mislead after early-return removal** _(confidence 0.62)_

Early-return removal for unknown redirect targets in _walk_redirected now appends an UNKNOWN decision and then walks the command node, which also triggers nested substitution collection. This can double-count reasons for a single physical statement: once for the redirect and again for the command/substitutions, producing concatenated reasons via aggregate_decisions. While decision precedence (unsafe > unknown > safe) is preserved, the joined reasons can be misleading (“writes to a file via redirection; terraform destroy: mutating” or multiple UNKNOWNs) and may alter downstream UX/log parsing or policy that expects one primary reason. The change increases the cardinality of reasons without adjusting aggregation semantics.

**Recommendation:** Harden aggregation or de-dup reasons at the redirected_statement layer: (a) tag reasons with a source key ({kind: redirect|command|substitution}) and collapse duplicates before aggregate_decisions; or (b) when adding an UNKNOWN for unreadable redirect, set a flag and suppress adding an identical UNKNOWN later; or (c) modify aggregate_decisions to uniquify reasons and prefer a single dominant reason per precedence band. Add tests that assert stable single-reason outputs for lines combining unreadable redirects with mutating commands.

hooks/grammar_classifier.py
```diff
@@ -167,12 +180,24 @@ def _walk_redirected(self, node, src, decisions, _depth):
                           unsafe_target)),
             ))
             return
-        if any(not self._target_is_safe_write(t) for t in write_targets):
+        if unreadable_target or any(
```
debedb: **[high] Unsafe-redirect early return masks command classification (asymmetry)** _(confidence 0.70)_

Asymmetry remains: unsafe redirect targets still return early, skipping command/node walk, while unknown targets fall through. A truly destructive command with a protected redirect target yields only the redirect-based UNSAFE reason, not the command’s own UNSAFE reason. This can mask the command classification and lose critical context for forensic/logging use cases and explainability. It also creates a precedence inversion at the reason level: redirect UNSAFE always wins and suppresses the command’s own UNSAFE details.

**Recommendation:** Make behavior consistent: do not return after unsafe redirect. Append the UNSAFE redirect reason, then continue to classify the command and aggregate; or, if returning is desired for performance, enrich the redirect UNSAFE reason with command context (e.g., include command name) and add a comment/test documenting the trade-off. Add a test asserting both redirect UNSAFE and command UNSAFE are represented or, at minimum, that command context survives.

tests/test_variable_write_targets.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def test_harmless_command_still_reports_the_redirect(self):
+        # The unknown verdict is still produced when nothing outranks it.
+        self.assertEqual(classify("echo hi > $LOG"), "unknown")
+
+    def test_protected_target_still_wins_over_a_safe_command(self):
+        self.assertEqual(
+            classify("echo x > $HOME/.ssh/authorized_keys"), "unsafe")
+
+
```
debedb: **[medium] Missing test coverage for UNSAFE redirect + mutating command coexistence** _(confidence 0.66)_

No explicit test covers the scenario where an unsafe redirect target coexists with a mutating command (e.g., `rm -rf / > $HOME/.bashrc`). The current asymmetry returns early and will never surface the command’s UNSAFE reason or verify aggregation behavior. This is a gap given the behavioral change rationale hinges on aggregation precedence; the unsafe side should be exercised too to catch masking or double-reporting.

**Recommendation:** Add a regression test: classify("rm -rf / > $HOME/.ssh/authorized_keys") asserts overall UNSAFE and that the reason includes both a redirect path and a mutating-command indication (or, if you decide to keep early return, asserts the chosen intended single reason and documents it). This forces an explicit policy and prevents future silent masking.

debedb (commented):

Do not ship yet. The change fixes one masking bug but introduces asymmetry and reason fan‑out risks. Unsafe redirect still short‑circuits, masking command classification and losing forensic context; unknown redirect now concatenates multiple reasons, which can mislead or break consumers. There’s also an untested recursion/visit increase surface. Align behavior, de‑dup reasons, and add targeted tests before release.

### Next steps
- Decide policy: keep or remove unsafe-redirect early return; implement consistent handling and write tests.
- Deduplicate or structure reasons to avoid misleading concatenation; adjust aggregate_decisions or callers.
- Add deep-nesting tests to ensure recursion guard isn’t triggered by fall-through; tune limits or visits if needed.
- Backfill tests for unsafe-redirect + mutating-command coexistence and for reason determinism on unknown-redirect + safe-command.

- `hooks/grammar_classifier.py:62-153` **[low] Increased recursion/visit count may trigger depth guard and weaken verdicts** _(confidence 0.45)_
  By falling through after adding an UNKNOWN redirect decision, the walker recurses into command and substitution nodes. In deeply nested shells with multiple redirections and substitutions, this increases the number of visited nodes relative to the prior early-return path. Combined with MAX_RECURSION_DEPTH=8 and multiple redirected_statement layers, there is a plausible path to hitting the recursion guard more often, flipping an otherwise classifiable line to DECISION_UNKNOWN ("max recursion depth"). While not an infinite loop, this is a performance/behavioral regression surface created by the change. _(fix: Add targeted tests with nested redirected_statement + command_substitution chains to ensure the recursion depth limit is not exceeded by the new fall-through. Consider counting and short‑circuiting only once per redirected_statement: after emitting the redirect UNKNOWN, skip walking into substitution-heavy branches if the command has already been classified at this depth. Alternatively, raise MAX_RECURSION_DEPTH or introduce a per-statement visit budget.)_
- `hooks/rule_classifier.py:448-461` **[medium] Reason-format compatibility regression via concatenation** _(confidence 0.55)_
  aggregate_decisions joins all reasons in a precedence band with '; '. After this change, statements with unknown redirect + unsafe command will report only UNSAFE reasons (as desired), but statements with unknown redirect + safe command now surface UNKNOWN with concatenated reasons. If prior consumers relied on a stable single-line reason (“writes to a file via redirection”), the appended command context may be unexpected and could weaken triage (harder to pattern-match prompts). This is a compatibility regression risk induced by reason fan‑out. _(fix: Stabilize reason format across fall-through cases: either (a) prefer the redirect UNKNOWN reason alone when it is present unless an UNSAFE also exists, or (b) introduce structured reason objects and render concise, deterministic primary+secondary messages. Add a test asserting reason determinism for ‘echo > $LOG’ remains a single redirect reason.)_

</details>

debedb:
**All three findings hold and are fixed in `5d8d662`.** Re-running was the
right call: r1 examined `accbb7c`, and `fb56935` — the masking fix written in
response to r1 — had never been reviewed. That is the second time on this
branch that the follow-up commit was the riskier one.

## The asymmetry (0.70) cost more than explainability

I left the unsafe-redirect early return in place deliberately, reasoning that
the verdict is `unsafe` either way so nothing is masked. Wrong, and the cost
lands somewhere I did not anticipate:

```
rm -rf /tmp/x > ~/.bashrc
  before: writes to protected path '~/.bashrc' via redirection
  after:  writes to protected path '~/.bashrc' via redirection; rm: mutating
```

`scripts/replay_unsafe.py` groups the corpus **by reason**, and #100 drafts
the non-delegable list from that grouping. A masked `rm: mutating`
under-counts `rm` in the measurement that decides what Phase 3 deletes. The
reason is data, not commentary — which is exactly the argument the finding
made, and I had not connected it to the gate.

Both branches now record and continue; the existing `unsafe > unknown > safe`
precedence picks the verdict.

## Reason duplication (0.62) — half right

Confirmed, but not in the shape described. The predicted
`"writes to a file via redirection; terraform destroy: mutating"` does not
occur — unsafe outranks unknown, so only unsafe reasons survive aggregation.
The real defect is repetition at one level:

```
echo a > $X; echo b > $Y
  before: writes to a file via redirection; writes to a file via redirection
  after:  writes to a file via redirection
```

De-duplicated, order preserved.

## Missing coverage (0.66) — legitimate either way

Correct that no test pinned the policy. Now pinned in both directions rather
than only the fixed side: a destructive command keeps its reason beside a
protected redirect, a **safe** command contributes none, and a repeated
reason appears once. Mutation-checked — restoring the early return fails two.

698 in the suite.

## On the review

Three findings, all three real, none refuted. That is a better hit rate than
either previous round on this PR, and the difference is that it was pointed
at a specific unreviewed commit rather than the whole diff.

Worth recording for `codex-adversarial-pr-review`: **re-review after a
review-driven fix, and scope `--focus` to the new commit.** On this branch
the follow-up commit was the riskier one twice running — once introducing a
regression, once leaving an asymmetry the original fix created.

